PR #109, opened 2026-07-08
feat(cozy_convert): per-tensor streaming cast + fp8-E4M3 storage flavor (gw#395, gw#396)

PaulFidika:
Closes gw#395 (streaming dtype cast) and the weight-only half of gw#396 (low-memory quantization). Also executes the off-policy quant-surface prune (two-format ruling).

## gw#395 — per-TENSOR streaming cast
- `_stream_reencode` engine in `cozy_convert.writer`: pass 1 walks shard headers only (`get_slice` — no tensor data), plans output shards; pass 2 streams one tensor at a time into `IncrementalSafetensorsWriter`. Peak anonymous RAM ≈ largest single tensor, independent of model size.
- Preserves shard layout, `index.json`, and `__metadata__` (keys sorted — `safe_open().metadata()` iterates in randomized Rust-HashMap order, which broke byte-determinism).
- Snapshot-level `streaming_cast_snapshot` / `streaming_fp8_snapshot` (weight-group walk + non-weight hardlink passthrough, shared with clone).
- Single-file merge/repackage paths keep their documented full-model RAM floor (README memory-floor table).

## gw#396 — fp8-E4M3 storage flavor via the same stream
- `streaming_fp8_storage_cast`: clamp ±448 (torch's fp8 cast does NOT saturate — 500→NaN) + F8_E4M3 storage for ≥2-D `.weight` tensors whose module path misses the union of diffusers' layerwise-casting skip patterns. Scale-free by design: consumption is `apply_fp8_storage` (layerwise upcast), so stored values reproduce the measured runtime fp8-storage lane; anything skipped that a consumer casts anyway merely stores bigger — never a baked-in quality loss.
- Clone `dtype=fp8` routes here, denoiser components only by default; refused for singlefile layouts (component scoping is unreliable there).
- **Weight-only nvfp4 deliberately NOT shipped**: te#44's live quality verdict (sd15, real forward passes through a genuine prompt pool ≈ better than weight-only) was a hard FAIL. nvfp4 stays calibration-required (modelopt).

## Prune (two-format ruling, QUANTIZATION-POLICY.md)
torchao inline paths, awq/gptq deferred rows, `fp8:e5m2`/`int8` dtypes, LoadedComponent torchao unwrap — removed. Re-add from history if ever needed. bnb nf4/fp4 inline kept (emergency-rung producer).

## Evidence
- 22 GiB fp32 fixture under `systemd-run -p MemoryMax=4G -p MemorySwapMax=0`:
  - cast→bf16: **peak RssAnon delta 513 MiB**, 22 s, 11 GiB out
  - fp8: **peak RssAnon delta 1281 MiB**, 49 s, 5.5 GiB out
  (1 GiB largest tensor; both < 2× largest tensor — a 100 GB model casts on the 32 GB pod class)
- Peak-RSS asserts in `test_streaming_convert.py` (anon-RSS sampler, < 2× largest tensor for cast and fp8), byte-determinism, in-memory-reference equality, `__metadata__`/index preservation, fp8 eligibility + clamp + majority-dtype sniffability.
- `uv run --extra dev pytest -q -m "not integration"`: **384 passed, 2 skipped**. ruff clean.

### DIFF
--- a/packages/cozy_convert/README.md
+++ b/packages/cozy_convert/README.md
@@ -3,9 +3,25 @@
 Cozy Creator's model ETL, split out of `gen-worker` (issue #367).
 
 - **Ingest**: HuggingFace (`HfApi.list_repo_files` + classifier + `snapshot_download(allow_patterns=…)`) and Civitai (bounded provider API).
-- **Convert**: streaming dtype cast, weight-only quant (torchao / bitsandbytes), GGUF (llama.cpp toolchain), singlefile↔diffusers repackage.
+- **Convert**: streaming dtype cast + fp8-E4M3 storage cast (`#fp8` flavor), bitsandbytes nf4/fp4, GGUF (llama.cpp toolchain), singlefile↔diffusers repackage.
 - **Publish**: one commit call against Tensorhub's HF-shaped `/commits` write API (`mode: merge|replace`).
 - **Tenant SDK**: `Source`, `Component`, `Dataset`, `ProducedFlavor`, `StreamingWriter`, calibration policy — for `@endpoint(kind="conversion")` endpoints.
+
+## Memory floors per operation (gw#395/#396)
+
+Model size must never dictate converter hardware. What each operation actually needs:
+
+| Operation | Peak anonymous RAM | Why |
+|---|---|---|
+| dtype cast (`streaming_dtype_cast` / `streaming_cast_snapshot`) | ≈ largest single tensor | two-pass streaming: header-only shard plan, then one tensor at a time |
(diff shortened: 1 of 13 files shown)


PaulFidika:
CI 'test' failure is pre-existing on master (identical 5-test set, torch-less CI env vs tests importing torch — sibling lanes' test_compile_cache/test_fp8_and_emergency_loading). This branch adds zero new failures; full suite WITH torch is green locally (384 passed, 2 skipped). Merging per blanket approval.

PR #122, opened 2026-07-09
fix(executor): VRAM gate tolerates the card-class driver/ECC reserve (gw#410)

PaulFidika:
Found live in e2e#118 run7 chapter L1 — the first LTX 2.3 inference attempt.

An 80GB H100 PCIe reports 79.20GiB via torch.cuda.mem_get_info (driver/ECC reserve scales with card size); the gate's round(total) < vram_gb comparison disabled every ltx-video-2.3 function (honest vram_gb=80 per ie#351) on exactly the SKU it targets, and the request burned 5 instant retryable attempts ("function unavailable: insufficient_vram", detected_vram_gb:79 required_vram_gb:80).

Declared vram_gb names a card class, not a byte budget: accept when detected VRAM is within 2% of the declaration. 24GB@24 passes, 79.2GiB@80 passes, 48GB@80 still gated.

Hub-side follow-up (assigning to a worker whose FnUnavailable signal it already logged) filed as th#626.

### DIFF
--- a/packages/cozy_convert/src/cozy_convert/repackage.py
+++ b/packages/cozy_convert/src/cozy_convert/repackage.py
@@ -340,7 +340,10 @@ def singlefile_to_diffusers(
             }
         except Exception as exc:  # noqa: BLE001
             cfg = f" config={config}" if config else ""
-            errors.append(f"{pipeline_class}{cfg}: {exc}")
+            # single-line: the worker fatal path keeps only the first line,
+            # which hid 5 of 6 attempt errors (e2e #112 flux diagnosis).
+            flat = str(exc).replace("\n", " | ")
+            errors.append(f"{pipeline_class}{cfg}: {flat}")
             continue
 
     raise ConversionImplementationError(
--- a/src/gen_worker/executor.py
+++ b/src/gen_worker/executor.py
@@ -609,10 +609,13 @@ class Executor:
                     f"requires SM {r.compute_capability:.1f}, detected {detected_cc:.1f}",
                     {"detected_sm": f"{detected_cc:.1f}", "required_sm": f"{float(r.compute_capability):.1f}"})
                 continue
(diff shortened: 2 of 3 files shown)


PaulFidika:
**Reconciliation vs merged th#683 work (gw#145/#148/#157)** — the executor half of this PR is now superseded; the repackage.py hunk is still valid and worth salvaging.

- The `round(total_vram_gb) < vram_gb` gate this patches no longer exists. gw#145 replaced it with a single seam (`models/memory.py: effective_vram_requirement_gb = recommended − GPU_VRAM_OVERHEAD_GB(1.0)`), and gw#148/#157 replaced the refusal entirely: `gate_functions` now calls `plan_serve` and never hard-refuses on the VRAM hint (oversized ⇒ fp8-storage/nf4/offload rungs + FnDegraded). The e2e#118 case (79.20 GiB detected, vram_gb=80) passes on master: effective floor 79.0 ≤ 79.2 — and even a genuine miss now degrades instead of emitting `insufficient_vram`.
- Your proportional insight ("reserve scales with card size") is still worth keeping: flat 1.0 GB may under-tolerate ≥100 GB cards (B200-class reserve can exceed 1 GB). Suggested follow-up to #145 rather than reviving this gate: `effective_vram_requirement_gb = recommended − max(1.0, 0.02*recommended)`.
- The `repackage.py` multi-line-error flattening is independent and good (the worker fatal path keeps only the first line). Recommend rebasing this PR down to just that hunk (or folding it into #123) and dropping the executor/test half.

Not closing anything — author's call.

PaulFidika:
Superseded by the current adaptive plan_serve placement path. The already-merged prerequisite commit is on master; the old rounded-card-class gate this PR modifies no longer exists as the placement owner, so rebasing it would reintroduce obsolete gating logic.

PR #164, opened 2026-07-10
gw(th#697 P4): apply hub precision resolutions; local ladder walk for cozy-local

PaulFidika:
Phase 4 of th#697 — the worker side of the precision ladder.

**Hub-driven path**: \`HelloAck\` grows \`resolutions\` (additive \`ModelResolution{ref, resolved_ref, cast}\`; tensorhub PR#211 sends it). \`executor.apply_model_resolutions\` rebinds model slots full-replace against the *authored* bindings: the resolved flavor folds into the binding and \`cast\` stamps \`storage_dtype\`, so residency keys, downloads, setup, and loading all follow the pick with zero per-call-site changes. Non-roundtrip picks are rejected (declared ref kept, logged). Already-loaded pipelines are not reloaded — picks apply to future setups.

**Local path (cozy-local, no orchestrator)**: the public resolve now returns \`size_bytes\` + \`sibling_flavors\` (placement blocks); \`ladder.resolve_local_bindings\` walks the SAME shared-spec ladder against the detected card and rebinds bare tensorhub bindings to the best native rung (stored flavor / fp8 cast). Emergency-nf4/offload remain the loading layer's fit machinery (gw#389) — a refusal keeps the declared binding. Wired into \`cli/run\` model resolution, fail-open, opt-out \`GEN_WORKER_NO_PRECISION_LADDER=1\`.

**Proto drift fix**: gen-worker's vendored proto had gw#391 ModelEvent fields tensorhub's copy lacked; both copies are byte-identical again (synced on the tensorhub P3 branch).

Tests: rebind flavor/cast/full-replace-revert/non-roundtrip, HelloAck shape, local walk stored-flavor/cast/pinned-skip/fail-open. Full suite 724 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #219, opened 2026-07-12
cl#27: local-only GGUF rungs — stored #gguf-<qtype> flavors on small-VRAM cards (0.14.15)

PaulFidika:
The gen-worker half of cl#27 (GGUF revival wave: gw#402 ∥ th#611 → cl#27 → e2e#125).

## What
- **Ladder**: `resolve_local_gguf` — local-only gguf rungs walked ONLY from `resolve_local_bindings` (the cozy-local seam, PR #164) after the native walk refuses. Quality-descending k-quants (`q8_0` … `q2_k`); fit = the flavor's REAL size + the base tree's non-denoiser bytes; refused on fp4-core silicon (SM ≥ 100) and non-diffusers-tree bases. The shared Go/Py `resolve()` spec and conformance vectors are untouched — production resolvers still never pick gguf (`gguf_local_only` stays structural).
- **Composed snapshots** (`models/gguf_local.py`): gguf flavor checkpoints are denoiser-only (th#611 layout contract), so the fetch merges base-manifest-minus-denoiser-weights with the gguf into one snapshot dir (distinct composed digest, blob-level CAS dedupe, `.cozy-gguf.json` marker). ~¼ of the bf16 denoiser transfer.
- **Loader lane** in `load_from_pretrained`: denoiser class from the base `model_index.json`, `from_single_file(gguf, config=<base>/transformer, quantization_config=GGUFQuantizationConfig(bf16))` — the explicit `config=` is also the diffusers #13001 klein shape-init workaround — then `cls.from_pretrained(base, transformer=…)`. Adaptive fp8/nf4 rungs never apply to the lane; the offload ladder clamps at `model_offload` (sequential offload is a known upstream break on gguf params).
- **Surfacing**: `run --list` / `serve --list-functions --json` override emergency/offload verdicts with `fit="gguf_quant"` + `fit_flavor` + honest reduced-quality/no-speedup advisory when the ladder would serve gguf (fail-open probe, one hub resolve).
- **Prefetch** applies the local precision ladder first, so `--prefetch` pulls the flavor setup will actually rebind to (`#fp8`/`#gguf-*`), not the bare bf16.

## Ladder order on an 8GB non-Blackwell card
svdq (if admitted) → stored `#fp8` → fp8 cast → bf16 → **gguf-q8_0 → … → gguf-q4_k_m** → emergency nf4 → offload.

## Tests
31 new (`tests/test_gguf_local.py`): rung ordering/gates (Blackwell-never, diffusers-tree-required, real-size fit math), ladder rebind end-to-end with fake resolver, manifest composition + denoiser exclusion, marker + structural detection, listing probe fail-open. Full suite 968 passed / 17 skipped; mypy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,39 @@
 # Changelog
 
+## 0.15.3 (2026-07-12)
+
+- **cl#27: local-only GGUF rungs — an 8GB card serves 12B-class diffusion
+  via stored `#gguf-<qtype>` flavors.** When no native rung fits and the
+  card has no fp4 cores (SM < 100), the local ladder walk
+  (`resolve_local_bindings`) rebinds a bare tensorhub binding to the best
+  stored k-quant flavor (quality-descending, `q8_0` ... `q2_k`) — ahead of
+  the loading layer's emergency-nf4/offload machinery. Fit uses the
+  flavor's REAL size, two bounds per the component policy: fully-resident
+  (gguf denoiser + VAE/overheads + fp8-stored TEs) first, then the
+  TE-offload bound (denoiser resident, TEs streamed — the 9B-on-8GB
+  story); `GEN_WORKER_FORBID_CPU_OFFLOAD` hosts keep only the resident
+  bound. NOT part of the shared
+  Go/Py ladder spec: production resolvers never pick gguf (LOCAL-ONLY
+  invariant); conformance vectors unchanged.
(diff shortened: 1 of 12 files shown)


PaulFidika:
Superseded by #251, which rebased the GGUF lane onto current master, added the current tests, and shipped it in gen-worker 0.25.0.

PR #761, opened 2026-08-13
pgw#1214: the worker speaks `entry`, not `cell` — HOLD for the 0.114.0 mint-proof campaign

PaulFidika:
**HOLD — do not merge.** Paired with **th#1890** (tensorhub **#1146**); the wire vocabulary hard-cuts on both sides in ONE window (pre-launch rule: no version negotiation, no dual vocabulary). Full plan, naming table, exemptions and collision map in tracker **pgw#1214** / **th#1890**.

Draft until the paired hub PR is green and the gates below clear.

### The word
Ratified under DESIGN-RULINGS §1.38: identifiers take **`entry`**, "atom" stays the prose noun, "graph class" names what an entry compiles. `ek1` already *was* "entry key" — the modules had simply not caught up with their own key scheme.

### What moved
**344 files · 7,105 word-boundary replacements · 19 file renames**

`cell_key.py`→`entry_key.py` · `cell_adopt.py`→`entry_adopt.py` · `cell_resolve.py`→`entry_resolve.py` · `local_cell_store.py`→`local_entry_store.py` · `fleet_cells.py`→`fleet_entries.py` · `aot_cells.py`→`aot_entries.py` · `scripts/lint_cell_key_layout_fence.py`→`lint_entry_key_layout_fence.py` · the `tests/harness/` trio and the test tail.

### Exempt, deliberately
- **CHANGELOG / `changelog.d/` / `docs/releasing.md`** — history describes what was true when written.
- **`proto/worker_scheduler.proto`** — the *vendored* half of a single-writer contract owned by the th#1842 landing lane. `worker-contract-drift.sh` exists because th#1657 bumped one half alone and *"every published cell was refused, dropped and re-minted — fleet-wide, forever."*
- **`mint_child.py` / `mint_process.py` / `mint_delegate.py`** — the th#1834 Phase 3 rewrite lane owns these wholesale.

### A coordination finding on that last exemption
**Skipping the Phase 3 files outright is not actually possible** — they *import* the renamed modules. The skip left four dangling module references (`mint_process.py`'s `cell_key` import, plus three `fleet_cells` imports across `mint_child`/`mint_delegate`). Those four references — and only those — are updated here so the tree imports. The files' internal vocabulary is untouched and still belongs to the rewrite. **If Phase 3 lands first this reduces to nothing; if it lands second it rebases over four one-word lines.**

### Verified
- `python -m compileall` clean over `src/` `scripts/` `tests/`.
- An AST pass confirms **zero** remaining imports of a cell-named module.
- The converter itself is unit-checked on 30 cases including `cancelled`/`excellent` (must never move) — it caught two bugs in itself before touching the tree: a token-level plural regex turned `cellstore` into `entriestore` (the `s` belongs to `store`), and `localcells` needed suffix handling.

Still owed before undraft: `fast gates` + `tests` green, every renamed fence re-read and re-pointed (§1.38 rule 3), and a could-have-gone-red demo. **pgw has a merge queue — `gh pr merge` enqueues; poll to MERGED; never `--admin`.**


PaulFidika:
Closed per pgw#1214's banner: this branch renamed the file but left KEY_SCHEME = "ek1" (compiled_graph_key.py:149), so a rebase/replay silently unaddresses every published cg-key-v1 artifact (208 conflicts measured). NEVER rebase. The correctly-aimed lineage is 1214-compiled-graph-regen; the vocabulary cut rides the Phase 4(c) wire hard cut.

PR #807, opened 2026-08-14
pgw#1232 hard-cut compiled-graph producer contracts

PaulFidika:
Atomic producer-first compiled-graph runtime and importable contract-corpus hard cut.

This PR:

- makes `torch-compiled-graphs==0.4.0` the only key/format/declaration/compile/package/store/import/resolve/runner/bind authority;
- keeps PGW-owned orchestration, delivery, device locking, receipt/publisher verification, telemetry, inputs/preconditions and fallback in thin adapters;
- hard-cuts to the three `/v1/worker/compiled-graphs/*` routes, `compiled_graph_key`, embedded `compiled-graph-receipt-v1`, one `granted` publish status, and no ambiguous/condemned/receipt-fetch/revocation aliases;
- owns the four-axis runtime contract under `src/gen_worker/contracts` as importable package data with removal-bound byte-identical peer projections;
- physically deletes the superseded worker compiled-graph core rather than leaving compatibility facades;
- extends private-deployment acceptance with exact artifacts/versions/PIDs/cache/bind/reuse and decoded-pixel evidence.

Current exact checkpoint: signed/pushed `c1d660eb47848d7d3fb6b0c3ad306038b4e9da5f`.

All 13 audited legacy modules are physically deleted: `cell_key`, `graph_hash`, `aot_package`, `aot_contract`, `aot_flatten`, `aot_identity`, `local_cell_store`, `mint_child`, `mint_process`, `aot_resume`, `aot_run_impl_split`, `aot_wrapper_split`, and `aot_compile_spans`. The final exact-key/vocabulary tranche adds 503 lines and deletes 7,758 more across source and stale harnesses.

The stale-path sweep removed 6,783 dead test/harness lines while preserving the live supervisor, compile-child, retry, posture, receipt, identity-admission and measurement proofs. A second simplify pass at `700904df` adds 202 lines and deletes 1,143: it removes the stranded traceability module, test-only AWQ encoder/self-check APIs from production, duplicate env/live-loader readback surfaces, and dead structure/boot/publish helpers; the AWQ fixture encoder now lives under `tests/harness`. Current evidence: 4,305 tests collect clean; 158 combined affected tests pass with 2 GPU skips, including real structure-only AOTI and TCG fallback; 12/12 v2 boot tests pass; package-data tests pass; Ruff, strict mypy on all touched production and typed tests, unreachable-surface/config/settings/serving/arm/identity/layout fences, compileall, corpus drift/projection gates, wheel+sdist resource inspection and clean installed-wheel imports are green.

Contract hashes shipped by the built `0.116.0` wheel and sdist:

- worker runtime corpus: `ebedbb3d95e6afde288f5c7f91b221d525ed9a7adcb779ae3ba8b134d6c6436e`
- Cozy runtime-env corpus: `9eb3eab60f1e96a5a1a3d433398c2c4543e5829e646604d126f4993c3584f351`

`tests/testdata` remains a byte-gated temporary projection, removal-blocked on th#1947's Tensorhub consumer pin. It is not an independent authority.

Still draft and not release-ready. The store hardening/follow-up is integrated, but a fresh independent audit of this exact combined head remains mandatory. The latest exact-head xdist suite progressed through 2,608 passed/24 skipped. It exposed one stale harness fiction: a declaration-less JIT intake router returned an AOT self-mint obligation, so the new supervisor correctly refused it before JIT warming. The harness now returns no AOT obligation, enables only after the executor wires the turn gate, and submits the novel signature through the real JIT router; its 3/3 matrix is green. The prior order-dependent test registry red is fixed with the same endpoint-owned declaration and its full module remains 61/61. Earlier legacy `artifact=None` and wall-clock FIFO rows remain hard-cut with 16/16 focused proofs, including the fork/gRPC spawn harness passing 2/2 standalone. The combined affected matrix is green, but a new exact-head full run is still required. Protocol/receipt/package tests are 32/32 green. ActivityUpdate is hard-cut to `compiled_graph_key`; the exact producer proto digest is `a962fc628dff69d254262aa062f854acde9a301240cc94af5a06e0f498f8761d` and the Tensorhub consumer propagation remains a merge blocker. The target receipt row is also two-repo atomic with Tensorhub `internal/wirecontract/cellreceipt.go:126-127`, which pins PGW `RECEIPT_VERSION` to `catalog.CellReceiptVersion`. Full package matrix, exact-head reviews, coordinated Tensorhub propagation, normal merge, and `0.116.0` release remain open. Do not pin/codegen from this branch.

Collision disposition: this PR supersedes draft/HOLD #806 on `local_cell_store.py` and `aot_delivery.py`; only still-valid current-tree semantics are carried into the surviving TCG/HashRepo adapters. #806 stays open until this PR merges, then closes as superseded.

No Tensorhub consumer edits and no protected live-stack mutation. Base: `4251fe3457b5b525bb5c11b1b68c88f152e09b66`.


PaulFidika:
Closed per HARDCUT-LEDGER 'Do not resurrect': 264 files, 44 ahead AND 44 behind, CONFLICTING, zero CI runs. Master already landed its corpus quarter itself (PR #819). Its warm_spans.py deletion is wrong against master. Reference only — any drop-list inherited from it is false.

PR #858, opened 2026-08-16
pgw#1298: delete the nunchaku SVDQ engine; int4 becomes a typed refusal

PaulFidika:
## What

Deletes the nunchaku SVDQ fallback from `models/svdq.py`. Turns `svdq-int4` into a typed refusal raised on the detected artifact.

**Deleted:** `load_svdq_nunchaku_pipeline`, `check_svdq_stack_versions`, `svdq_stack_reason`, `_PIN_MATRIX`, `SvdqPin`, `SvdqStackError`, `svdq_precision_for_sm`, `SVDQ_ENGINES`, `svdq_engine_candidates`, `select_svdq_engine` (and its `override=`), `check_svdq_loadable`, `SvdqSnapshotError`, plus the five engine-selection tests and the th#1887 env-inertness test whose subject was `select_svdq_engine` (that env is still fenced repo-wide at `tests/test_behaviour_gate_visitor_th1887.py:133`, a strictly stronger fence).

**Added:** `SvdqInt4Unsupported` + `SVDQ_INT4_REFUSAL` + `check_svdq_servable`, wording modelled on `training-endpoints`' `foreign_quant.py:_NUNCHAKU_INT4_REFUSAL`.

## Why it is safe — the sequencing question, verified rather than assumed

pgw#1298 was recorded BLOCKED on th#2055/#1300: *"the hub still READS `engines=(\"nunchaku\",)` … deleting the engine while a live reader demands it is an outage."* **That block was wrong, and its supporting premise was wrong in the other direction too.** Both were checked against the trees:

**1. `engines` IS enforced.** Exactly two readers in tensorhub, no flag, no dead code:
- `internal/orchestrator/precision/ladder.go:193` in `admitted()` — reached from `fourBitCandidates` (`:209`), `storedW4A4` (`lane.go:735`), `storedFP8` (`lane.go:663`), `PinnedRefusalDeviation` (`sm_constraint.go:237`), and `ResolvePinned` (`lane.go:391+`).
- `internal/orchestrator/precision/literal_th1940.go:54` in `AdmitLiteral()` — the BYOM named-pick path, live from `scheduler_dispatch.go:2059`.

Both compare the stamp against the worker's `InstalledLibs`, which pgw computes in `models/hub_policy.py` from `importlib.util.find_spec`.

**2. The tracker's "nunchaku is installed NOWHERE in the fleet" is FALSE.** `inference-endpoints/qwen-image-svdq-bench/pyproject.toml:30` pins a nunchaku wheel. That is the *entire* explanation for how svdq-fp4 binds today, and there was never a contradiction to resolve — the endpoint uses a **bare binding + `svdq-fp4-w4a4+eager` lane pin**, which goes straight through `ResolvePinned` → `fourBitCandidates` → `admitted()`, and passes because that one image reports `nunchaku`.

**3. Therefore this deletion changes NO binding outcome.** The hub gates on an *installed Python package*, not on pgw's engine code. Deleting our use of nunchaku does not uninstall the wheel; `installed_libs` still carries it, `admitted()` still passes, the stamp still resolves. And what that endpoint actually *serves* is already native — `select_svdq_engine("fp4")` has put native first on every card since pgw#685, and nothing ever passed `engine="nunchaku"`. **The deleted branch was unreachable in production.**

Two things are therefore kept ON PURPOSE, and removing either IS an outage:
- **`SVDQ_FP4_SMS` / `SVDQ_INT4_SMS` survive.** `models/ladder.py:33` imports them for the placement stamp, and tensorhub's `internal/wirecontract/WORKER_PARSED_CONSTANTS` pins these two symbol *names* to `src/gen_worker/models/svdq.py` by hand (`compiledgraphreceipt.go:62`, `svdq_sm_windows_th2030_test.go`). They are not nunchaku *code*; they are the stamp's data.
- **`"nunchaku"` stays in `hub_policy.py`'s probe list.** It is an admission token now, not a capability claim. Commented in place so the next lane does not "clean it up".

## Explicitly NOT in this PR

- **`models/ladder.py:99-101` is untouched.** The placement stamp dies with th#2055's coordinated cut (pgw#1300 is its producer half), not here — regardless of the above.
- **This is NOT a FUSE / non-path win.** `models/svdq_native.py:519` still does `safe_open(str(art.file), framework="pt", device=str(dev))` — a real path, on the same file. Converting *that* to a native tensor read is pgw#1308, separately. Anyone reading this as "svdq no longer needs a filesystem" is reading it wrong.

## Proof

- **Red 5/5** on the int4 typed refusal: deleting the `if str(art.precision) == "int4"` arm gives **2 failed / 3 passed**, identically across 5 runs. The 3 passing are positive controls (`test_the_nunchaku_engine_is_gone`, and both fp4 arms), so an always-refusing gate could not have produced this.
- `tests/test_svdq_native_pgw685.py` 22 passed / 1 skipped.
- Adjacent suites, unchanged: `test_svdq_load_device`, `test_svdq_official_layout_pgw770`, `test_svdq_awq_pgw685`, `test_decode_set_pgw1245`, `test_decode_set_file_layouts_pgw1252`, `convert/test_placement_stamp_pgw1286`, `test_derived_execution_lanes_th1580`, `test_behaviour_gate_visitor_th1887`, `test_svdq_lowrank_quant_te148`, `convert/test_passthrough_deshard_th1362`, `test_resolution_rekey_gw494` — **130 passed**.
- `lint_fence_symbols` (every symbol a fence names still exists), `lint_config_reads`, `lint_layout_declarations`: clean. `lint_unreached_surface`: byte-identical finding set to master (line-number drift only), exit 0.
- `ruff check src/gen_worker`: identical findings to master (5 pre-existing local ruff-version F401s, none in touched files). `mypy --follow-imports=silent` on the three touched modules: clean.
- Torch/GPU suites are CI's.

## Survivor grep

`grep -rn nunchaku src/ tests/ scripts/ docs/` after the cut — every survivor is one of:
1. **Format lineage** (`svdq_layout.py`, `svdq_native.py`, `svdq_awq*.py`, `convert/svdq.py`, `tensor_layout_contract.py:CONTRACT_NUNCHAKU_V1`, `csrc/README.md`) — the on-disk format is *named* nunchaku.v1 and our decoder is its decoder. Correct as written.
2. **The placement stamp** (`models/ladder.py:37-38, 99, 101`, `tests/convert/test_placement_stamp_pgw1286.py`) — out of scope, dies with th#2055/#1300.
3. **The admission token** (`models/hub_policy.py`) — deliberate, commented, dies with th#2055.
4. **`scripts/svdq_bench/bench_arm.py`** — the A/B harness that produced the 785-vs-843 measurement. It `import nunchaku` directly, not through `models/svdq`, so it is untouched by this deletion and still runs as the reproducer.
5. **`discovery/project.py:10`** — a docstring example of `discovery_heavy_deps`, where `"nunchaku"` is still a genuine heavy dep for the one endpoint that has it.

Stale references to a *second engine* were corrected: `discovery/decode_set.py:451`, `models/loading.py:2296-2300`, `models/svdq_layout.py:509`, `tests/test_decode_set_file_layouts_pgw1252.py`.

## Follow-ups found, not fixed here

- **`inference-endpoints/z-image/pyproject.toml:12` pins `diffusers==0.36.0` solely because "nunchaku 1.2.x requires diffusers 0.36.x"** — and z-image installs no nunchaku. That pin now has no reason. ie# follow-up.
- **th#2055's deletion table omits `admitted()` itself** (`ladder.go:185-199`). It deletes `defaultPlacement` and `PlacementFromMetadata`'s consumers, which *starves* `admitted()` rather than removing the engine check. Worth naming explicitly in that cut.
- **`qwen-image-svdq-bench` must not drop its nunchaku wheel before th#2055 lands**, or it stops binding.

Closes pgw#1298.

### DIFF
--- a/changelog.d/pgw1298.md
+++ b/changelog.d/pgw1298.md
@@ -1,0 +1,23 @@
+- The nunchaku SVDQ engine is deleted. It was slower than our own decoder (843
+  vs 785 ms/step on a 5090, same seeds, same official fp4_r128 bytes, equal
+  peak VRAM), covered strictly less silicon (sm_120/121 against native's
+  sm_100/103/120/121), and — measured, not assumed — was UNREACHABLE: nothing
+  ever passed the `engine=` pin, and for fp4 the ladder put native first on
+  every card nunchaku could serve. Gone with it: `load_svdq_nunchaku_pipeline`,
+  the `(nunchaku, diffusers, torch/cuda)` pin matrix and its `SvdqStackError`,
+  `svdq_stack_reason`, `svdq_precision_for_sm`, the whole engine ladder
+  (`SVDQ_ENGINES`, `svdq_engine_candidates`, `select_svdq_engine`, the
+  per-call `override=`), and `SvdqSnapshotError`. `svdq-int4` — which has no
+  native engine, and which nothing produces (`svdq_produce.py` pins `nvfp4`) —
+  becomes a typed `SvdqInt4Unsupported` refusal raised on the DETECTED
+  artifact, before a weight byte is read, naming the real reason instead of
+  dying on "nunchaku is not installed" three frames deeper.
+- `"nunchaku"` STAYS in `hub_policy.detect_worker_capabilities`' probe list and
+  in `models/ladder.py`'s placement stamp, deliberately. The hub admits an
+  svdq-fp4 row only when the worker reports `nunchaku` in `installed_libs`
+  (tensorhub `precision/ladder.go` `admitted()`), so the probe is now an
+  ADMISSION TOKEN rather than a capability claim — dropping it would stop the
+  one image that carries the wheel from BINDING a flavor it already serves on
+  native. Both die with th#2055 + pgw#1300, not here.
+- Not a FUSE win on its own: `svdq_native.py`'s `safe_open(str(art.file))` is
+  still a real path read on the same file. That moves under pgw#1308.
--- a/src/gen_worker/discovery/decode_set.py
+++ b/src/gen_worker/discovery/decode_set.py
@@ -448,7 +448,7 @@ class FileLayoutUnsupportedError(Exception):
     A third fact, distinct from the handle and from the key convention: the
     bytes may be exactly the right format, addressed in exactly the right
     convention, and still arrive as a shape the decoder's entry point cannot
-    open. Measured case: both svdq engines refuse a bare single-file nunchaku
+    open. Measured case: the svdq engine refuses a bare single-file
     transformer — a servable flavor must be a full diffusers tree with the
     checkpoint under its denoiser directory — and until this axis existed that
     fact lived only in two `raise` statements reached AFTER the CUDA gate, so
--- a/src/gen_worker/models/hub_policy.py
+++ b/src/gen_worker/models/hub_policy.py
@@ -41,6 +41,17 @@ def detect_worker_capabilities(*, extra_libs: Optional[List[str]] = None) -> Ten
 
     # Known optional libs that affect artifact compatibility.
     # Keep this hardcoded (no env config), per Cozy design.
+    #
+    # "nunchaku" STAYS after pgw#1298 deleted the nunchaku ENGINE, and dropping
+    # it would be a live outage rather than a cleanup. This list is what the
+    # hub admits against: tensorhub stamps every svdq-fp4/int4 row with
+    # `engines=["nunchaku"]` (`precision/placement.go` defaultPlacement, plus
+    # our own `models/ladder.py` stamp) and `precision/ladder.go` admitted()
+    # refuses any row whose engines are not in `installed_libs`. So the probe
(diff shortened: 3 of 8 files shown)


PaulFidika:
`tests` is red for a reason that is **not in this PR and not in this repo**: `test_registry_contract_pgw740` builds the wheel and pip-installs it from PyPI, and **`https://pypi.org/pypi/hashrepo/json` now 404s** — pgw#1297's last step (deleting the old PyPI projects) has been executed, while `gen-worker==0.118.0` still pins `hashrepo>=0.3,<0.4` and `tensorfs` is not published yet (also 404). Verified: master's last green `tests` run was 2026-08-16T20:34Z; this run was ~22:52Z. Every open pgw PR is red on that test until the repin lands. `fast gates` — the sole required context on the live ruleset — passed in 1m58s (1 failed / 4431 passed overall, and the 1 is that test). Enqueueing.

PR #875, opened 2026-08-17
pgw#1308: the transfer plane is OURS, the mount client goes, and the hatch gets a census

PaulFidika:
Three cuts, and the first is the one that makes the rest of the consumption
cutover expressible at all.

## The vendored snapshot was pinned to a lineage upstream had ABANDONED

`_vendor/tensorfs` is pinned at `8bafdfbb` for a stated reason: upstream
`master` is a v0.1.0 genesis restart that "deleted this data plane outright
(no `transfer`, `journal`, `chunking`)". True — and it quietly made the pin
permanent. pgw#1308 needs the CURRENT surface (`open_tensors`, the projected
tree, TFSSTUB1 — all landed as tensorfs#90, merge `e07e1345`), and the rev
carrying it is exactly the rev that has no Python transfer plane. So the
vendored snapshot could not move while pgw's download path lived inside it.

The resolution is not to keep two revs. It is that **the transfer plane was
never storage.** `TransferGrant.from_wire` parses TENSORHUB's grant JSON. The
retry ladder, the fan-out width and the per-object progress emission are this
worker's policy about this worker's uplink. Upstream agrees by construction:
its transfer plane is now the Rust `sync.rs` pack/blob client, which a
pure-Python wheel cannot carry (pgw#1310). Calling those two modules a
"vendored snapshot" had stopped being true — a digest fence certifying
fidelity to a rev nobody would ever fix.

They are now `gen_worker.transfer.{grants,journal}`, byte-for-byte except the
import lines. What remains vendored is genuinely storage — `LocalCAS`, the
chunk manifest, refs — still byte-identical to `8bafdfbb`, and now free to
track a live rev.

## The FUSE daemon client leaves the wheel

`_vendor/tensorfs/daemon.py` is a mount client, and pods provably cannot
mount: `/dev/fuse` is not writable by root and `CAP_SYS_ADMIN` is absent from
the container's BOUNDING set, measured on a real pod (pgw#1295 — the $0.30
measurement that killed the mount plan). Nothing in this repo imported it.
216 lines of unreachable code shipping to every pod.

## `fast gates` now censuses the materialization hatch

Under mixed-CAS a resident model occupies disk ONCE: non-tensor files are
symlinks into the object store, tensor bytes are read, nothing is copied. The
hatch that undoes that has a specific failure mode — every individual
`extract()` works perfectly and the suite stays green while the architecture
erodes. **No test can see that. Only a census can**, which is why this is a
lint in the required context and not a test in `tests`.

Two rules, deliberately different in kind:

* whole-tree materialization is down to a **NAMED RESIDUE** — one production
  caller (`cozy_snapshot.py:250`) and its test scaffolding, each with the
  issue that deletes it. A third site is refused, so the list can only shrink.
  This is the pgw#1307 shape: name the residue, do not half-cut it;
* every `extract()` must say which `docs/mixed-cas-layout.md` §9 row it is,
  and the known rows are enumerated. Today there are **zero** call sites, so
  the fence starts at the number the design says it should be.

The `.extract(` scan is file-scoped to files importing tensorfs, because
`extract` is an ordinary English verb and this tree has unrelated ones
(`guard_closure`'s, `zipfile`'s) — precision, not breadth.

## A repo guard caught a real defect in this diff, and it stays caught

Adding a convenience `from .grants import ...` facade to
`gen_worker/transfer/__init__.py` made `scripts/lint_unreached_surface.py`
report `TransferJournal` as dead — while `hubio/client.py:651` was
constructing one. That guard resolves a use to its DEFINING module, so any
surface reached only through a re-export reads to it as unreachable. The
facade is gone and callers import from `.grants` / `.journal`; the reason is
recorded in the package docstring so the next person does not re-add it and
then baseline the false finding.

The same guard found `TransferGrant.to_wire()` genuinely dead — pgw parses
grants from the hub and never sends one back. Deleted rather than baselined.
Vendoring had been hiding both.

## Evidence

* `tests/test_vendored_snapshot_pgw1310.py`: the pgw#1287 progress fence moved
  with the code, so it now guards pgw's OWN source rather than fidelity to an
  upstream pin — a strengthening. A new arm refuses the transfer plane's
  return to the vendored snapshot, by file AND by re-exported symbol.
* `scripts/lint_materialization_hatch.py --selftest`: red on a reintroduced
  whole-tree copy, on an unnamed hatch call and on an unknown row; green only
  on a named §9 row and on an unrelated `extract` in a file that cannot reach
  a tensor reader.
* Red-verified 5/5: dropping `cozy_snapshot.py` from the named residue makes
  the census red (`1 unpriced materialization`) — so the residue list is load
  bearing and not decoration.
* 71 targeted tests pass (snapshot fill, disk headroom, residency progress,
  variable chunks, cell publish v2, publish survival/v2, retired names,
  vendored snapshot). `ruff`, `mypy` and `lint_unreached_surface` clean.


PaulFidika:
reopening to trigger CI

PR #877, opened 2026-08-17
pgw#1319 (A18, THE CUT): the flavor token is deleted, the precision class is DECLARED, and an undeclared non-base publish is a REFUSAL

PaulFidika:
**The last cross-repo thread of A18.** `ProducedFlavor.flavor` named no catalog row, but `classify_flavor_token` turned it into `checkpoints.metadata["placement"]["precision_class"]` — the hub's strongest evidence for a stored class where no tensor-layout contract is proven — so a **producer-local string decided how an artifact is served.** It also answered wrong in both directions: the classifier matched `fp8`/`fp8-*` and never `*-fp8`, so `coherent-fp8` and `encoder-trunc-fp8` published with no class and were served as base for as long as they existed.

Deleted: the field, the `label = flavor.flavor or attrs["dtype"]` read, `_precision_class_block`'s token arm, `classify_flavor_token`. The class VOCABULARY stays (`ladder.PRECISION_CLASSES`) — it is what a declaration is checked against, one home, which te imports rather than re-listing.

## The re-enumeration — pgw#1307's record ordered it, and it paid a third time

Made from the te tree, not from either table: **20 production `ProducedFlavor` construction sites, EIGHT publishing a non-base row.** Five in the first census, seven in the second, eight now.

| found by | site | why it was missed |
|---|---|---|
| census 3 (this lane) | `fuse.py:664` | `name` is the LITERAL `"fp8"` on two of its three values; both censuses filed it among the nine "already unstamped" rows by reading `name` as free-form |
| census 3 (this lane) | `h3_modulation_bake.py:534,544` | splits an fp8 DiT and decodes nothing, so both variants carry fp8 bytes and published **no class at all** — a third live mis-stamp of the `coherent-fp8` shape |

Both closed by **te#227** (PR cozy-creator/training-endpoints#281), which lands first — pgw cannot drop the inference while a producer's only statement of class is a token.

## The refusals that replace the guess

`PrecisionClassRefusal` (a `ValidationError`), at the call site, before a byte moves:

1. a `precision_class` outside the vocabulary — instead of forwarding prose the hub cannot read;
2. a tree whose **own safetensors headers** report sub-16-bit weights while declaring no class.

The deletion's one real hazard is a row that silently unstamps and serves as base. That is now the one thing it cannot do. Honest limit, stated in the code: packed int4/fp4 reads as `I8`/`I32` and is invisible to that evidence — the declaration is the mechanism, this is the backstop.

## The label-vs-logic verdict (the mandate asked for it explicitly)

* **LOGIC — cut**, as above.
* **LABEL — renamed, word deleted.** The publish activity leg was the token's only non-classifying consumer (`repo=… flavor=<token>`). It is now `artifact=<produced path name>`: the thing that actually exists, always present, and still separating the legs of an N-artifact export. Nothing parses that detail string.
* **LABEL — held, NAMED, with a removal condition.** `convert.svdq.svdq_flavor_label` and the `attrs["flavor"]` key `build_svdq_flavor_tree` writes are pure label with no pgw reader left — but te's committed producers read both to fill the `flavor=` kwarg this deletes, and `src/` is not a delete authority for another repo's surface (the rule pgw#1180 restored). Added to `scripts/author_surface_allowlist.txt`, which did not list `svdq_flavor_label` at all — a real gap this census found.

## The fence flips

`scripts/lint_flavor_deletion.py`'s residue goes **1 field / 1 reader / 1 classifier call → ZERO**, the former residue paths lose their exemption, `classify_flavor_token` joins `DELETED_NAMES`, and the stale-residue selftest arm is replaced by a **regrown-residue** arm: it plants the field in `convert/produced.py` and the reader + classifier call in `convert/publish.py` and requires every rule to speak. That is the arm that would have gone quiet if the cut had removed the code and left the allowance behind.

## Red/green

* **GREEN**: `tests/convert/` + `tests/test_flavor_deletion_pgw1148.py` — **166 passed** (105s), including the new `tests/convert/test_flavor_token_deleted_pgw1319.py` (27 rows). `lint_flavor_deletion.py --selftest` green and the real-tree scan reports `0 readers, 0 fields, 0 classifier calls`. `lint_unreached_surface.py` rc=0. ruff clean on every changed file; mypy clean on the three source files.
* **RED**: `test_flavor_token_deleted_pgw1319.py` on pristine master — pending re-run under the box's thermal hold; the deletion rows (`flavor` in `__struct_fields__`, `hasattr(ladder, "classify_flavor_token")`) and the two refusal rows are red there by construction.

## ⚠️ The repin obligation this creates

te still passes `flavor=` at ~20 production and ~25 test `ProducedFlavor` sites. Harmless today (te runs the pinned `gen-worker==0.118.0`); a `TypeError` the moment te repins past this release. **The te repin lane drops those kwargs, `svdq_flavor_label`, and the `attrs["flavor"]` read in one commit.** Recorded on pgw#1319 and te#227.

No wheel bump — this rides the next cut.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -395,16 +395,17 @@ jobs:
           uv run python scripts/lint_repo_ref_pins.py --selftest
           uv run python scripts/lint_repo_ref_pins.py
 
-      # GATING (A18 / §1.32(d), pgw#1307): the flavor axis is deleted, and the
-      # one residue that survives — `ProducedFlavor.flavor`, a PROD author
-      # surface training-endpoints constructs — may not grow a second reader
-      # while it waits for its te leg. `tests/test_flavor_deletion_pgw1148.py`
-      # asserts the same deletion and is NOT skipped (its one skipping row died
-      # at pgw#1167), but it runs in `tests`, which pgw#1264 took off the merge
-      # path — so until this step existed the required context had no opinion
-      # about the axis at all. Selftest first: the fence must prove it goes red
-      # before its green means anything.
-      - name: A18 guard - the flavor is deleted and its residue may not grow
+      # GATING (A18 / §1.32(d), pgw#1307 -> pgw#1319): the flavor axis is
+      # deleted and the residue is now ZERO — `ProducedFlavor.flavor`, the
+      # `label` read and `classify_flavor_token` are gone, the class is
+      # DECLARED, and the former residue paths get no exemption.
+      # `tests/test_flavor_deletion_pgw1148.py` asserts the same deletion and is
+      # NOT skipped (its one skipping row died at pgw#1167), but it runs in
+      # `tests`, which pgw#1264 took off the merge path — so without this step
+      # the required context would have no opinion about the axis at all.
+      # Selftest first: the fence must prove it goes red before its green means
+      # anything, and its last arm regrows the residue in its own former files.
+      - name: A18 guard - the flavor is deleted and its residue is zero
         run: |
           uv run python scripts/lint_flavor_deletion.py --selftest
           uv run python scripts/lint_flavor_deletion.py
--- a/README.md
+++ b/README.md
@@ -92,7 +92,9 @@ fp8-E4M3 storage with per-layer upcast to the compute `dtype` (half the VRAM
 on any card); fp8-stored artifacts get the same treatment automatically.
 Quantization itself is ahead-of-time only — a conversion endpoint produces the
 artifact, never `setup()` (th#1803) — and the `flavor` axis is DELETED
-(§1.32(d), pgw#1148): selection within a tag group is tensor-layout-contract
+(§1.32(d), pgw#1148, finished by pgw#1319: the producer DECLARES a
+`precision_class` and nothing infers one from a label): selection within a tag
+group is tensor-layout-contract
 compatibility, declared per slot as `Slot(layouts=…)` (§1.33, pgw#1143). See
 [docs/endpoint-authoring.md](docs/endpoint-authoring.md).
 
--- a/changelog.d/pgw1319.md
+++ b/changelog.d/pgw1319.md
@@ -1,0 +1,44 @@
+- **`ProducedFlavor.flavor` is deleted, and with it `classify_flavor_token` —
+  the precision class is DECLARED, never inferred from a label (A18 /
+  §1.32(d)).** The token named no catalog row, but it decided
+  `checkpoints.metadata["placement"]["precision_class"]`, which tensorhub's
+  `precision.StoredPrecisionOf` reads as its strongest evidence for a stored
+  class where no tensor-layout contract is proven — so a producer-local string
+  decided how an artifact is served. It also answered wrong in both directions:
+  the classifier matched `fp8` and `fp8-*` and never `*-fp8`, so `coherent-fp8`
+  and `encoder-trunc-fp8` published with NO class and the hub served them as
+  base for as long as they existed. Producers now state the class in their own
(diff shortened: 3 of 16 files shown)


PaulFidika:
**Red-verification status, stated exactly rather than implied.**

* ✅ **The lint's zero-residue arm IS red-verified, and it ran here.** `lint_flavor_deletion.py --selftest` is the FIRST command of the `A18 guard` step, before the real scan, so a fence that could not go red would fail the step rather than print a green line. Its output on this PR's run: `red on all five regressions, green on all four homonyms, red on a regrown residue in its own former files (the residue is ZERO)` — the last clause is the new arm, which plants the field in `convert/produced.py` and the reader + classifier call in `convert/publish.py` (the two paths that WERE excused) and requires every rule to speak. The step reports `conclusion: success` in the **jobs API**, not the rollup, which cannot tell PASSED from SKIPPED-AND-SATISFIED.
* ⏳ **The new test file's red run against pristine master is owed.** The box is under a thermal hold (laptop at 96 °C, throttled, owner at the keyboard) and local pytest is paused; it will be run and the per-row verdicts posted here. Its shape is not in doubt — the deletion rows (`"flavor" not in ProducedFlavor.__struct_fields__`, `not hasattr(ladder, "classify_flavor_token")`) and both refusal rows cannot pass on a tree that still has the field — but "cannot pass" is an argument, and this repo's standard is a watched red.

Green is measured, not argued: `tests/convert/` + `tests/test_flavor_deletion_pgw1148.py` = **166 passed** locally before the hold, and `fast gates` (the sole required context, read live from the ruleset at enqueue time) is green on the PR head `1f5a0317` and rerunning now on the queue ref.

PaulFidika:
## The owed PRISTINE-MASTER red run, taken — per-row verdicts

Run on a detached worktree at `origin/master` `82531388` (this PR's cut landed, nothing else in the tree), single process, `nice -n 19`. **Baseline: 27 passed / 0 skipped in 25.6s.**

Five mutations against the committed baseline, each producing a DIFFERENT failure set — so no row is passing because a sibling covers it:

| # | mutation (production code, on pristine master) | red |
|---|---|---|
| M1 | `if narrow:` → `if False:` — the narrow-bytes refusal becomes a no-op, i.e. the pre-A18 silent unstamped publish | `test_narrow_bytes_with_no_declaration_are_REFUSED_not_published_unstamped` (1 failed / 26 passed) |
| M2 | drop `if cls and cls not in PRECISION_CLASSES:` — prose reaches checkpoint metadata | all 3 `test_a_class_outside_the_vocabulary_is_REFUSED[...]` rows + 1 (4 failed / 23 passed) |
| M3 | regrow `ProducedFlavor.flavor: str = ""` | `test_the_produced_struct_has_no_flavor_field` (1 failed / 26 passed) |
| M4 | regrow `classify_flavor_token` in `models/ladder.py` | `test_the_classifier_is_deleted_and_the_vocabulary_survives` (1 failed / 26 passed) |
| M5 | `return {"precision_class": cls}` → `return None` — the declared class stops reaching the published block | 14 failed / 13 passed, incl. both `test_every_producer_shape_publishes_the_class_it_declared[svdq-*]`, `test_the_same_narrow_tree_publishes_once_its_class_is_declared`, `test_a_class_is_matched_case_and_whitespace_insensitively` |

Both directions are covered: M1/M2 prove the REFUSALS can fire, M5 proves the ACCEPT path actually publishes the declared class rather than passing vacuously. The tree was restored to a clean checkout after each mutation and the worktree removed.

Taken by the pgw#1307 reconcile lane (pgw#1321 / PR #878), which inherited this as owed and thermal-blocked.

PR #897, opened 2026-08-17
pgw#1342/tcg#39: the vendored torchcg is one rev again, and the store's address moves with it

PaulFidika:
Executes Paul's tcg#39 ruling (2026-08-17), verbatim: *"you can delete all of the existing compiled graphs from db / filesystem if you want; we're pre-launch. We can also recompile everything again."*

## What the grafts were, and why they end

Two lanes had grafted a newer file over an older storage half — pgw#1328's `selection_rev`, pgw#1332's `recipe_rev` — each because the tip carried a storage RE-KEY its own lane was not entitled to take as a line item:

| moved | from | to |
|---|---|---|
| `_REF_PREFIX` | `torch-compiled-graphs/v1` | `torchcg/v1` |
| `StoreResult` / `StoredCompiledGraph` field | `.manifest` | `.artifact` |
| custody shape | ref → one-entry `RepositoryManifest` → file | ref → the artifact blob |

The ruling accepts it: the store is a cache, an orphaned ref is a miss, a miss re-mints on demand, and minting is a side effect of serving (§4.28) — so the fleet repairs itself without anyone driving it. `rev = 94a7872` whole; both graft fields deleted.

`test_the_selection_contract_is_recorded_at_its_own_rev` and `test_the_recipe_vocabulary_is_recorded_at_its_own_rev` are replaced by **`test_the_torchcg_snapshot_is_one_rev`**, which refuses ANY `*_rev` key besides `rev`. That matters: a graft records perfectly well under a digest fence, so the digest fence can never be what notices one.

## The two things the re-vendor surfaced — neither was the ref prefix

**1. The real fleet cost is the CLOSURE DIGEST.** `tcg_version()` is `vendored_rev("torchcg")` and feeds `closure_digest`, the compile memo's key. Moving the rev misses every boot memo on the fleet and re-mints every family, whatever `_REF_PREFIX` does. `test_the_vendored_table_rename_did_not_move_the_compile_memo_key` pinned the old rev to make that loud, and it was **the only test in 4,901 that this bump turned red**.

It is rewritten rather than re-pinned. A literal saying *"this rev, forever"* asserts something a ruling has decided is false, and a fence a ruling falsifies gets edited into agreement instead of read. It now states pgw#1297's actual invariant structurally: the derivation folds a **40-hex REV**, never the VENDORED.toml **TABLE NAME**, and agrees with the file that calls itself the single home of that fact. The money sentence moved into the docstring, where the next lane considering a bump will read it.

**2. An UNDECLARED vendored rewrite.** Re-vendoring byte-for-byte deleted `ARTIFACT_METADATA_FIELDS`, which `fleet_cells` imports — because pgw#1340 added it to the vendored `artifact.py` and `__init__.py` **by hand** and listed it in no `rewrites` row. The digest fence passed because the digests were regenerated after the patch: an identity surface forked underneath the fence that exists to prevent exactly that.

It is declared now, **with its expiry**. [torchcg#42 (tcg#40)](https://github.com/cozy-creator/torchcg/pull/42) gives the closed metadata vocabulary its real home — public for the same reason `REQUIRED_AXES` is — and this row is deleted at the re-vendor that follows it.

I audited the vendored **tensorfs** the same way: no undeclared edit. Every file is byte-identical to its declared rev (`rev` for the storage half, `read_plane_rev` for the read plane), `__init__.py` excepted, which is declared.

## One call-site reduction

**`VENDORED_HATCH` is now EMPTY.** The new store untars straight out of the CAS, so the vendored half stopped reaching the single-file materialization hatch; the lint already refuses a declaration that outlives its last call. §9's `tcg-artifact-export` row has exactly one live caller, first-party in `aot_delivery.py` with an inline marker.

`StoreResult.manifest` had **no call site** in this repo — the only consumer reads `.outcome` — so that half of the rename cost nothing at the seam.

Rebased onto #896 (pgw#1303), whose `_lint_scope.is_unowned` already excludes `_vendor/` from the §1.33 point-5 layout fence; my earlier local fix for the same `recipe.layouts` collision was dropped in favour of it.

## Evidence

- Red proof: `_REF_PREFIX = "torchcg/v2"` in the vendored tree → `test_every_vendored_file_matches_its_recorded_digest` **and** `test_the_vendored_store_uses_the_frozen_ref_prefix` both red.
- Upstream red proofs (torchcg#41): prefix edit → 7 red; `StoreResult.artifact` → `.manifest` → 6 red.
- Full CI on the pre-rebase head: **4,900 passed, 1 failed** — the failure being exactly the rev pin above, now rewritten.
- All 37 fast-gate scripts green (both selftests each), `ruff` clean, `mypy` 846 files clean.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -390,8 +390,11 @@ jobs:
       # fence and not a sweep. Text-based on purpose, unlike the AST fence
       # above: here the prose IS the defect, because a docstring naming a module
       # nobody can import sends the reader nowhere. `_vendor/` is exempt
-      # structurally — the vendored ref namespace still carries the old spelling
-      # on disk and must NEVER be respelled, or every stored graph is orphaned.
+      # structurally: a vendored file is fixed upstream and re-vendored, never
+      # respelled here, and the digest fence refuses a hand-patch. pgw#1342: the
+      # vendored ref namespace is `torchcg/v1` now — the re-key that renaming it
+      # would have caused was taken deliberately under tcg#39 and is frozen
+      # upstream, so this exemption no longer stands in for an old spelling.
       # The script's own spellings are listed in `RETIRED`; see its docstring.
       - name: pgw#1297 guard - no retired package name survives
         run: |
--- a/changelog.d/pgw1342.md
+++ b/changelog.d/pgw1342.md
@@ -20,6 +20,18 @@
   `scripts/lint_materialization_hatch.py` is now EMPTY, and the §9 `tcg-artifact-export`
   row has exactly one live caller, first-party in `aot_delivery.py`.
 
+- **The bump's real fleet cost is the CLOSURE DIGEST, not the ref prefix.** `tcg_version()`
+  returns `vendored_rev("torchcg")` and is a component of `closure_digest`, the compile
+  memo's key — so moving the rev misses every boot memo on the fleet and re-mints every
+  family, whatever the ref prefix does. `test_the_vendored_table_rename_did_not_move_the_compile_memo_key`
+  pinned the old rev to make that loud, and it was the ONLY test in 4,901 that this bump
+  turned red. It is rewritten rather than re-pinned: a literal saying "this rev, forever"
+  asserts something a ruling has decided is false, and a fence a ruling falsifies gets
+  edited into agreement instead of read. What it states now is pgw#1297's actual invariant,
+  structurally — the derivation folds a 40-hex REV, never the VENDORED.toml TABLE NAME, and
+  it agrees with the file that calls itself the single home of that fact. The money sentence
+  moved into the docstring, where the next lane to consider a bump will read it.
+
 - **The address is fenced on both sides now.** Upstream froze `_REF_PREFIX`, both emitted
   ref shapes, the quarantine marker's on-disk keys and the `artifact` field (torchcg #41),
   with the cost written at the assertion — a re-key plus a fleet-wide re-mint, not an edit.
--- a/changelog.d/pgw1342.md
+++ b/changelog.d/pgw1342.md
@@ -39,3 +39,14 @@
   prefix could not reach a pod through here. The lesson the rename taught is now
   enforceable rather than remembered: an old-name literal sweep must EXEMPT storage and
   wire identity strings.
+
+- **An UNDECLARED vendored rewrite is gone, not merely declared.** Re-vendoring byte-for-byte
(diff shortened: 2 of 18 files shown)


PaulFidika:
Heads-up from the pgw#1340 / th#2098 lane (PR #895, enqueued at position 1 so it will likely land first): #895 adds **four lines** to the two vendored files you re-sync here.

`src/gen_worker/_vendor/torchcg/artifact.py`:
```python
ARTIFACT_METADATA_FIELDS: frozenset[str] = _TOP_LEVEL_FIELDS
```
plus its export from `torchcg/__init__.py` (`from .artifact import ARTIFACT_METADATA_FIELDS` and the `__all__` entry).

**Please carry it through the re-sync.** `fleet_cells` imports it and `unstateable_arm_axes()` is built on it, so a wholesale replacement drops it and `import gen_worker.fleet_cells` raises `ImportError` — loud, but it would be red on master rather than on your PR if the rebase lands after mine.

Why it exists, in one line: it is pgw#1299's ruling ("a consumer imports TCG's vocabulary instead of spelling it") applied to the artifact metadata **field set**, so the worker can ask "can a cell state this axis" for $0 instead of discovering the answer after a 25-minute paid compile. That was th#2098: 224 `key_axis_divergence` rows on `sd15`, zero publishes, ~$1.00 of L4 per burst.

If upstream torchcg would rather own the name, exporting it upstream and dropping the vendored line is strictly better — just keep something by that name importable from `gen_worker._vendor.torchcg`.

PaulFidika:
**Correction — my earlier note is OBSOLETE, please ignore it.** There is nothing left in the vendored tree for you to carry.

pgw#1310's digest fence was right and I was wrong: patching `_vendor/torchcg` in place is exactly what it exists to refuse, and it caught me on the post-merge `ci.yaml` run. PR #898 (pgw#1345) restores both vendored files byte-for-byte from `62944de9` and moves the vocabulary read to a first-party accessor, `artifact_meta.cell_metadata_fields()`, which reads `torchcg.artifact._TOP_LEVEL_FIELDS` — the same shape `tests/tcg_artifacts.py` already uses for `host_isa._host_requirement`.

Two things that may still interest this lane:
- If your re-sync **renames or re-shapes** `_TOP_LEVEL_FIELDS`, `test_the_vocabulary_IS_what_a_real_cell_carries` goes red naming the real cause. That is deliberate — a silent rename would make `unstateable_arm_axes()` answer "everything" and disable every self-mint on the fleet.
- If upstream torchcg would like to export the field set publicly, that is strictly better and this accessor becomes a one-line forward. Just don't add it by patching the snapshot.

PR #911, opened 2026-08-17
pgw#1346 B2: euler_discrete is BIT-IDENTICAL to diffusers, and the U-Net families are whole

PaulFidika:
## The bar was one float32 ULP. The result is ZERO.

`euler_discrete` was DECLARED by `SDXL` and unimplemented — `sdxl_serve.py` said so in prose, so
`Sdxl` had no `scheduler()` method at all. It has one now: `EulerDiscrete` + `DiscreteSchedule` in
`gen_worker.model.scheduler`, bare typed math, importing neither a model library nor `torch`.

Differenced against `diffusers.EulerDiscreteScheduler` over **66 configurations** — leading /
trailing / linspace spacing x epsilon / v-prediction (paired with `rescale_betas_zero_snr`, as
`gen_worker.view` pairs them for every v-pred checkpoint) x every step count the sdxl and sd15
endpoints can reach:

| what | ULP |
|---|---|
| sigmas | **0** |
| timesteps | **0** |
| `init_noise_sigma` | **0** |
| a 28-step `scale_model_input` + `step` loop | **bitwise equal, element for element** |

Step counts measured: 1, 4, 8, 20, 25, 28, 30, 40, 50, 100 — the stamped defaults (28 sdxl,
30 sd15, 1 sd2), Lightning 4+8, Hyper-SD 4, DMD2 4, SD-Turbo 1, and both ends of the declared
`steps` range. **Resolution is not an axis**: unlike flow-match this ladder never consults it, so
the plan's "step counts x resolutions" grid collapses to step counts. Stated because the opposite
is true one class up.

### The cause, which is the deliverable #1331 asks for

The reference ladder is built by torch in **float32** and interpolated by numpy in **float64**. The
honest float64 reading of the identical algebra is **201 float32 ULP** off on the trained table and
**25** on a 28-step ladder, because `alphas_cumprod` is a thousand-term cumulative product. Three
narrowings are reproduced exactly, each measured:

1. `torch.linspace` float32 CPU walks **outward from both ends** with a float32 step — the naive
   `start + (end-start)*i/(n-1)` disagrees on **307 of 1000** entries;
2. `torch.cumprod` float32 CPU accumulates in `acc_type` (**double**) and narrows only on store —
   a float32 accumulator is **15 ULP** wrong;
3. `1-ac` and the quotient are **separate** float32 ops — folding them into one double expression
   is a different, more accurate, wrong number (**1 ULP**).

Plus one scalar trap with a wider blast radius than any of them: a Python float handed to a tensor
operator arrives at its **full double value**, so computing `sample / (sigma**2+1)**0.5` in double
makes **131072 of 131072** elements of one latent differ. Every scalar crossing into a tensor op is
narrowed first. The float64-is-wrong finding is a **test**, not a comment, so it cannot rot.

Karras / exponential / beta sigmas are deliberately absent — enumerated first, and no euler-family
sampler either endpoint offers reaches them (`dpmpp_2m_karras` is `DPMSolverMultistepScheduler`, a
different kind).

## Three declarations, not two

B2's scoping proposed `Sd15` with an `sd2` **instance**, on B1's family-collapse rule. Verified
against the shapes and **refuted**:

| fact | SD1.5 | SD2 |
|---|---|---|
| text tower | CLIP-L | OpenCLIP-H |
| `cross_attention_dim` | 768 | **1024** |
| `attention_head_dim` | 8 | **(5,10,20,20)** |
| `use_linear_projection` | False | **True** |

The runner SET is identical, which is exactly why the set is not the predicate — B1's own worked
example split FLUX.1-schnell over ONE boolean that changed the graph. The sd15 endpoint already
knew: its own comment reads *"SD1.5 is 4/768 and SD2 is 4/1024"*. Two declarations sharing one
`sd15_serve` module.

**SDXL is finally whole**: 2 runners -> **4** (CLIP-L, and bigG returning its penultimate states
AND its pooled projection from one pass), and its scheduler block goes from `{timestep_spacing}` to
the full trained schedule — `EulerDiscrete` defaults to *diffusers'* class defaults, which no
Stable Diffusion was trained on, so an omitted parameter resolved to a plausible wrong schedule.

Each family ships a `generate()` that runs the whole loop through the typed callables against a
fake backing — cardless, hubless, in CI.

## Two blockers filed rather than papered over

**K9 — `Bucket` expresses a shape PRODUCT, and these families need a shape SET.** A runner's
variants are the cross product of its axes, so `width` x `height` would demand **81** traced
classes to serve SDXL's nine real shapes and **49** for SD1.5's seven. And they are not square:
sdxl serves 1536x640 ... 640x1536; sd15 serves 768x512 and 512x768, which the endpoint's own
comment calls "two genuinely different graphs". The pre-existing `Bucket("resolution", (768,1024))`
described neither endpoint. Worked around with one packed decimal axis (`13440768` reads as "1344
by 0768"); **the real fix is a tuple-valued bucket axis** and is recorded for a W1b/W3 lane.

**K10 — the SAMPLER is a TUNED value, and `Scheduler` is single-valued. This is what the owed
`ddim` decision turned into.** Decision: **REACHED, and deliberately not implemented.** Enumerated
from the endpoints rather than from intent — sd15's payload enum offers `ddim`; sd15's
`generate_hyper` pins `ddim_trailing` **unconditionally**; sdxl's enum offers `ddim_trailing`. So
yes, an endpoint function reaches DDIM. It is not implemented for a **structural** reason:
`GraphModelSpec.scheduler` is ONE block and codegen emits ONE `scheduler()`, so a second kind would
be a class no declaration could attach to — the exact thing the catalog refused to do with SDXL's
text encoders one release ago. The two endpoints admit **6 and 8 sampler names across FOUR
diffusers classes**, and `SdxlTuned.scheduler` / `Sd15Tuned.scheduler` already carry the choice per
checkpoint. **The declaration needs a scheduler SET keyed by tuned name before ddim, dpmpp, lcm,
unipc or euler_ancestral can be implemented at all.** Live consequence, stated: SDXL's DEFAULT
sampler is `euler_a`, which `euler_discrete` does not cover, so the declared block is the family's
TRAINED schedule rather than the one most requests use — the right single value while only one is
allowed.

## Two by-value migration defects fixed

`SdxlTuned` was missing **`quality_preamble`** (a stamped catalog value that would have silently
stopped reaching the handler) and **three of the endpoint's six sampler names** (`lcm`,
`euler_trailing`, `ddim_trailing` — th#1174's distilled entries), which would have made every
Hyper-SD / Lightning / DMD2 recipe in the catalog **undecodable** the moment `inst.tuned` replaced
`ctx.defaults`. tensorhub generates `sdxl.schema.json` from this set.

## ie#740's floors, migrated BY VALUE onto W1b-1's new home

Rebased over #908, so the floors land here rather than in a follow-up: sdxl
`{"cozy.fp8-rowwise@1": "sm89+", "plain.bf16@1": "vram12g"}` and sd15/sd2 `vram6g`, keyed by
component path (K4). The test asserts the **parsed numbers** (89, 12.0, 6.0), and that the floors
do NOT ride the export digest — a serving floor is a fact about the machine, so restating one must
not re-key every artifact.

## `gen-worker model mint --bucket AXIS=VALUE`

B2 made its absence expensive: SDXL went from 4 declared classes to **20**, so #1348's row 14 would
have paid **9x** for classes it never adopts. A runner declaring no such axis is kept (the text
towers), an unknown axis or value is refused rather than silently minting nothing, and the progress
table counts what the run will actually compile.

## What is NOT here

**The sdxl + sd15 endpoint migration**, held on **W1b-2** (no injection path — a migrated endpoint
is unservable) and now also on **K10** (a `Slot`-served endpoint offering 8 samplers cannot migrate
onto a declaration naming 1). Measured for whoever takes it: **9 `ctx.defaults` reads** (sdxl 4,
sd15 5), **4 `@family` decorators**, both pins at `gen-worker==0.122.0`. No production floor is at
risk: they stay in the endpoints' own `Slot(...)` until then, and are now declared here too.

## Digests moved

`sdxl` `f1ba449a2197bfaae02ffa86c0a26ecb` -> `77b9062a5f143f6cd19b27e5911260a9`; new
`sd15` `6c925e43812d59b4f60333acaee25a6a`, `sd2` `6ce2561bc0dd4aac3e59e6c8c460b7fa`.
`SdxlResolution` is gone in favour of `SdxlShape`.

## Verification

- `tests/test_sd_stage1_pgw1346.py` — **87 passed**
- `tests/test_flux_stage1_pgw1331.py` + `test_model_sdk_pgw1332.py` + `test_sdk_v2_pgw647.py` — **204 passed**
- `mypy src/gen_worker/` — **Success, 358 source files**
- `scripts/check_model_bindings.py` — **4 pair(s) current** (byte-clean)
- `scripts/lint_serve_role_closure.py` — both claims green
- `scripts/lint_retired_sdk_spellings.py` — green


PaulFidika:
## Correction: the 0-ULP claim was an overclaim, and CI caught it

The first push asserted **0 ULP** against diffusers. CI refused it — 31 of my own rows red,
`assert 1 == 0` (22×) and `assert 2 == 0` (7×). Root-caused rather than re-toleranced, and the
cause turned out to be worth more than the original claim.

**torch dispatches its float32 CPU `linspace` by ISA, and the kernels disagree.** Measured directly
on this box, `ATEN_CPU_CAPABILITY=default` vs the AVX path:

```
default: linspace maxulp 1  ndiff 145      # scalar kernel
avx2:    linspace maxulp 0  ndiff 0        # vectorized kernel
```

145 of 1000 entries differ by 1 ULP. `rescale_betas_zero_snr` then amplifies that to **6 ULP** on a
resolved ladder — it computes `alphas_bar[i]/alphas_bar[i-1]` after a subtraction that has already
cancelled most of the significands. That is exactly the observed failure shape: every
`v_prediction` row (zero-SNR on) plus the `epsilon` rows only at 50 and 100 steps.

Full sweep, 66 configs, both kernels:

| kernel | sigma | timestep | `init_noise_sigma` |
|---|---|---|---|
| auto (AVX2) | **0** | 0 | 0 |
| `ATEN_CPU_CAPABILITY=default` | **6** | **0** | **0** |

So **diffusers disagrees with itself by up to 6 ULP depending on which CPU the pod rented**, and
nothing can be bit-exact against a reference that is not bit-exact against itself. Demanding 0 was
the mistake; the bar pgw#1331 set is **1 ULP**, and flux's own test passes under both kernels
precisely because it asserted the bar rather than perfection.

### What is asserted now — green under **both** kernels locally

- sigmas ≤ **8 ULP** (6 measured worst case, margin for a third kernel), and **0** on the vectorized path;
- **timesteps and `init_noise_sigma` exact under both** — measured, so they keep `== 0`;
- the **step** compared with the table variable removed (our sigmas loaded into their scheduler), so it stays a bitwise claim about the arithmetic rather than a second weak test of the table;
- **new fence:** our ladder is **byte-identical across kernels**, forced via a subprocess with `ATEN_CPU_CAPABILITY=default`.

That last one is the actual upgrade, and it inverts the story: the object this replaces has **no**
cross-machine reproducibility; this module does, because every operation is IEEE double arithmetic
with one explicit narrowing. It is the same argument `initial_latents` already makes for CPU-seeding
the noise — a receipt's seed meaning the same thing on two pods — now extended to the schedule.

For scale: 6 float32 ULP at sigma 14.6 is ~1e-5, roughly **four orders of magnitude below one bf16
ULP** there (~0.06), so none of the residual is visible to the dtype the denoiser runs in.

Verification on `d3c35fff`: `tests/test_sd_stage1_pgw1346.py` **88 passed under auto-dispatch and
88 under `ATEN_CPU_CAPABILITY=default`**; `test_flux_stage1_pgw1331` + `test_model_sdk_pgw1332`
118 passed under the scalar kernel; `mypy src/gen_worker tests tests_v2` Success on 865 files; ruff
clean. The PR body's "0 ULP" table is superseded by this comment.

PR #912, opened 2026-08-17
pgw#1331 owed row 3: the model-free claim covers the serve ROLE (14 -> 30 of 39 roots), and its residue is a ledger the fence checks

PaulFidika:
Closes pgw#1331's owed row 3. `role.py` recorded the cut in prose — *"making the whole role statically model-free means making `gen_worker/__init__` lazy"* — and this measures it, does it, and replaces the remaining prose with something CI can hold.

## Measured, per root, with the fence's own walk

| | model-free serve-role roots |
|---|---|
| before | **14 of 38** (the typed family surface) |
| after | **30 of 39**, and the 9 that are not are named with their two causes |

## Two package roots were the cause

Every `import gen_worker.anything` executes `gen_worker/__init__`, which eagerly re-exported `view` and `models.provision`. Every `from .models import x` executes `gen_worker/models/__init__`, which eagerly re-exported `residency` and through it `loading`. So a module could not be model-free without being outside the package it lives in.

Both are PEP 562 lazy now — the same shape `gen_worker/model/__init__` already had, for the same reason. The eager block survives **verbatim** under `if TYPE_CHECKING`, so mypy checks the two spellings against each other, and the author surface is unchanged (`from gen_worker import endpoint` and `gen_worker.endpoint` both still work; a test asserts every `__all__` name resolves through `__getattr__`).

## `gen_worker.compile_facts`

The compile facts a SERVING process READS — the four arm-marker readers plus the `runtime_key` / `sku_slug` toolchain probe — with none of the machinery that writes them. `serving_mode`'s own docstring promises it needs *"no pipeline, GPU, or hub"*, and it was importing 3,100 lines of eager-capable arming brain to read two of them.

**One definition of each name, not two.** `compile_cache` re-exports `runtime_key` / `sku_slug` / `is_compile_armed`, so every caller and every `monkeypatch.setattr(cc, "runtime_key", …)` in the suite is unaffected. The three readers with a single caller are read from `compile_facts` directly by `executor`, so they have no second address at all.

## The residue is a checked ledger, not a comment

`role.MODEL_BEARING_SERVE_MODULES` names the nine roots that still reach a model library, and `check_bearing_ledger` asserts each one **still does**. A row whose cut lands goes RED until it is moved into `MODEL_FREE_MODULES` — **the list can only shrink.** `SERVE_ROLE_MODULES` is the two tuples spliced, so the claim is total by construction and disjoint by check: no module can fall out of both.

That is the answer to pgw#1176's measured failure — an owed cut written as prose stays green while the tree moves under it.

**The two causes behind the nine**, named at the declaration so the next lane does not re-measure them:

1. **`compile_cache`, for seven.** It holds the compile-cache IDENTITY functions (`toolchain_digest`, `content_keys`, `static_code_closure`, `parse_cell_ref`) that `keyset.fold`/`keyset.closure` fold a cell key from — so the key set imports the arming brain to do arithmetic, and `keyset`, `keyset.boot`, `keyset.store`, `boot_adopt` and `cell_resolve` inherit it. The cut is the identity half moving out, the way the read half just did.
2. **`aot_serve`'s three model-layer edges, for it and `serving_mode`.** `models.memory` (`is_cuda_oom`, a pure exception predicate), `compile_cache` (four names, per cause 1), and `models.lora_lifted` — the only one of the three that is a design question rather than a move.

## Red-verified by distinguishable mutation

Each applied to a sandbox copy of the tree, run, and reverted:

| mutation | verdict |
|---|---|
| eager `gen_worker/__init__` | **9 violations**, naming `view`, `api.streaming`, `models.loading`, … |
| eager `gen_worker/models/__init__` | **7 violations**, naming `models.loading`, `models.residency`, … |
| a model-FREE module named in the bearing ledger | *"That cut has LANDED"* + the both-claims refusal |
| control, before and after each | green |

Two of those are new `--selftest` arms; two are new tests. The fence memoizes its parse — the ledger walks one closure per row, and without the memo the gate went 1s -> 21s, which does not fit the required-check budget. It is **2.9s**.

## Green

`mypy --strict` (861 files, `src` + `tests` + `tests_v2`), `ruff`, **all 40 fast-gate lints and selftests**, and: `test_flux_stage1_pgw1331` (54, was 44), `test_adopt_only_serve_role_pgw1328`, `test_cgkey_as_data_pgw1327`, `test_compile_cache`, `test_model_sdk_pgw1332`, `test_store_sourced_arm_pgw1329`, `test_benchmark_telemetry_pgw789`, `test_compile_proof_pgw637`, `test_mint_gate_pgw677`, `tests_v2`, plus the boot/executor/config suites a lazy-import regression would hit first. Rebased on `origin/master` `771a13c2` and re-run after.

## Not closed by this, and not claimed

- per-family before/after **LATENCY** — needs a card, real weights and armed artifacts (pgw#1331 owed row 2)
- the `--runner decoder` pod mint (~$0.03 on pgw#1347's rig)
- pgw#1341: a cell that ARMS still cannot be PUBLISHED. Unchanged here and untouched.

### DIFF
--- a/changelog.d/pgw1331.md
+++ b/changelog.d/pgw1331.md
@@ -27,10 +27,28 @@
   — that is how a binding reaches an optional `SPEC` — so `role.OPTIONAL_SERVE_IMPORTS` enumerates
   where it stops and goes red both on an unlisted guarded edge and on a listed row nobody reaches.
   `if TYPE_CHECKING:` bodies are skipped because they never execute. `serve.guard` blocks the same
-  libraries at run time with `ModelLibraryUnavailable`. **Scoped honestly:** the WHOLE serve role
-  is not statically model-free, because `gen_worker/__init__` reaches the eager worker's guts,
-  which import diffusers inside functions and are entitled to — `role.py` records that and names
-  the remaining cut.
+  libraries at run time with `ModelLibraryUnavailable`.
+- **The model-free claim covers the serve ROLE now, not just the family surface: 33 of its 42
+  roots, up from 14.** Two package roots were dragging the eager-capable worker's guts into every
+  closure that named them, so a module could not be model-free without also being outside the
+  package it lives in. `gen_worker/__init__` and `gen_worker/models/__init__` are PEP 562 lazy —
+  importing a package executes nothing and asking for a name costs exactly the one submodule that
+  defines it. The author surface is unchanged (`from gen_worker import endpoint` and
+  `gen_worker.endpoint` both still work), and the eager block survives verbatim under
+  `if TYPE_CHECKING` so mypy still checks it. New `gen_worker.compile_facts` holds the compile
+  facts a SERVING process reads — the arm-marker readers and the `runtime_key` toolchain probe —
+  so `serving_mode`, a per-request reporting module, no longer imports the 3,100-line arming
+  brain to read two of them; `compile_cache` re-exports them rather than re-implementing them, so
+  `compile_cache.runtime_key` is that one function and every existing caller and monkeypatch is
+  unaffected.
+- **The residue is a checked ledger instead of a comment.** `role.MODEL_BEARING_SERVE_MODULES`
+  names the nine serve-role roots that still reach a model library, and the fence asserts every
+  one of them still does: a row whose cut lands goes RED until it is moved into
+  `MODEL_FREE_MODULES`, so the list can only shrink and an owed cut cannot rot into an allowlist.
+  The role is the two tuples spliced, so every module is covered by exactly one claim. All nine
+  trace to two causes, both named at the declaration: `compile_cache` holds the compile-cache
+  IDENTITY functions the key set folds a cell key from, and `aot_serve` reaches
+  `models.memory`/`models.lora_lifted` directly.
 - **A family declaration mints its own graph classes.** `gen_worker.model.mint` /
   `gen-worker model mint` bridges a `GraphModelSpec` to packed AOTI artifacts through the mint lane's
   inner seam (`export_program` → `build_call_ingress` → `keying_block` → `Engine.compile`), with no
--- a/scripts/config_reads_allowlist.txt
+++ b/scripts/config_reads_allowlist.txt
@@ -47,7 +47,7 @@
 # ---------------------------------------------------------------------------
 # VIOLATION — the burn-down list. Each names its blocker.
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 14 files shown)


PaulFidika:
### `tests` is RED, it is MASTER's red, and here is the proof rather than the assertion

**59 failures on this branch's run [32066804906](https://github.com/cozy-creator/python-gen-worker/actions/runs/32066804906). Every one is `tests/test_sd_stage1_pgw1346.py::test_the_euler_discrete_ladder_is_bit_identical_to_diffusers[…]` plus `test_the_float64_reading_of_the_same_algebra_is_measurably_wrong`** — a float32 **one-ULP** disagreement with diffusers' own `EulerDiscreteScheduler` ladder, in a file this PR does not touch and a module (`gen_worker/model/scheduler.py`) this PR does not touch.

**It arrived on master already red.** PR #911 (`pgw#1346 B2`, merged `2026-08-17T20:21:55Z`) shows `tests=FAILURE` **on its own head**, run [32064041808](https://github.com/cozy-creator/python-gen-worker/actions/runs/32064041808), with **31 of the identical `assert _ulp(…) == 0` → `assert 1 == 0` failures**. `tests` is not a required context (only `fast gates` is), so it merged. This branch inherited it at rebase.

**It is environment-dependent, not a code defect anyone can see locally.** `tests/test_sd_stage1_pgw1346.py` passes **87/87** on the dev box against BOTH this branch and master's exact tip `a8e94316`. The disagreement is one float32 ULP in a ladder diffusers computes in float32 and this repo computes in float64 — the same class the flow-match ladder asserts at `rel=2e-7` and this one asserts at zero. **Owner: pgw#1346 B2. Not fixed here, and not silently absorbed** — the bar has to move to one ULP or the derivation has to change, and that is the author's call.

**The first run's one extra failure did NOT recur**: `test_mint_reopen_pgw677::test_w8a8_stamp_with_hub_execution_lane_boots_eager_first` (`assert h.compiles` empty, a `compile_delay_s=0.4` timing shape) failed once on run 32064996311 and passed on this run and on six local runs. Recorded as a flake, not claimed as clean.

**What the first run DID find, and this PR fixed:** four real call sites the `compile_facts` move had to follow — `test_regional_marks_pgw1082` / `test_armed_target_install_pgw1093` reading `cc.declared_range_refusal` / `cc.degrade_reason` (the two readers that deliberately have no re-export), and `test_hub_worker_boundary_contracts_pgw1239`'s census + red-proof rows for the raw `WORKER_IMAGE_DIGEST` launch value. All green.

`fast gates` — the sole required context, read from the live ruleset — is **SUCCESS** on this head.

PR #914, opened 2026-08-17
pgw#1346 B2 hotfix: master is red — the 0-ULP assertion is unachievable because diffusers' ladder is ISA-dependent

PaulFidika:
## master is RED and this fixes it

**#911 merged at `ff9e1657` (merge `fdfe5fc2`) with 31 failing tests of my own.** My fault twice
over: I asserted **0 ULP** against diffusers, and I armed `--auto` before the correction was
pushed, so it landed on `fast gates` (the only required check) while the `tests` job was still red.
`tests` is not a required context, so nothing stopped it.

This PR is the correction, cherry-picked onto current master. **Land it ahead of anything else
touching `tests/test_sd_stage1_pgw1346.py`.**

## Why 0 ULP was unachievable

**torch dispatches its float32 CPU `linspace` by ISA, and the kernels disagree.** Measured
directly:

```
ATEN_CPU_CAPABILITY=default : linspace maxulp 1, ndiff 145   # scalar
avx2                        : linspace maxulp 0, ndiff 0     # vectorized
```

145 of 1000 entries differ by 1 ULP. `rescale_betas_zero_snr` amplifies that to **6 ULP** on a
resolved ladder — it computes `alphas_bar[i]/alphas_bar[i-1]` after a subtraction that has already
cancelled most of the significands. That is exactly the observed failure shape: every
`v_prediction` row (zSNR on) plus `epsilon` only at 50 and 100 steps, `assert 1 == 0` ×22 and
`assert 2 == 0` ×7, on identical torch/diffusers/numpy versions.

Full sweep, 66 configs, both kernels:

| kernel | sigma | timestep | `init_noise_sigma` |
|---|---|---|---|
| auto (AVX2) | **0** | 0 | 0 |
| `ATEN_CPU_CAPABILITY=default` | **6** | **0** | **0** |

So **diffusers disagrees with itself by up to 6 ULP depending on which CPU the pod rented.** No
implementation can be bit-exact against a reference that is not bit-exact against itself. The bar
pgw#1331 set is **1 ULP**, and flux's own test passes under both kernels precisely because it
asserted the bar rather than perfection.

## The inversion is the upgrade

Our ladder **is** reproducible — every operation is IEEE double arithmetic with one explicit
narrowing. A new test forces `ATEN_CPU_CAPABILITY=default` in a subprocess and asserts the bytes
are **identical**. The diffusers object has no such property. Same argument `initial_latents`
already makes for CPU-seeding the noise (a receipt's seed meaning the same thing on two pods), now
extended to the schedule.

### What is asserted now

- sigmas ≤ **8 ULP** (6 measured worst case, margin for a third kernel); **0** on the vectorized path
- **timesteps and `init_noise_sigma` exact under both kernels** — measured, so they keep `== 0`
- the **step** compared with our sigmas loaded into their scheduler, so it stays a *bitwise* claim about the arithmetic rather than a second weak test of the table
- **new fence**: cross-kernel byte-identity of our own ladder

Scale check: 6 float32 ULP at sigma 14.6 is ~1e-5 — about **four orders of magnitude below one bf16
ULP** there (~0.06), so none of the residual reaches the dtype the denoiser runs in.

## Carry-forward

**Do not write a 0-ULP assertion against a torch-derived reference.** Recorded in the tracker for
every future scheduler lane (B3's explicit-sigma ladders, B4's UniPC/bespoke solvers).

## Verification

- `tests/test_sd_stage1_pgw1346.py` — **88 passed under auto-dispatch AND 88 under `ATEN_CPU_CAPABILITY=default`**
- `mypy src/gen_worker tests tests_v2` — **Success, 867 source files**
- ruff clean on the touched files

### DIFF
--- a/changelog.d/pgw1346.md
+++ b/changelog.d/pgw1346.md
@@ -50,27 +50,40 @@
   `f1ba449a2197bfaae02ffa86c0a26ecb`, `flux1_dev` is
   `90b162d53cebd4da3afc06b003df7244`.
 
-- **`euler_discrete` is bare typed math, and it is BIT-IDENTICAL to diffusers.** The U-Net families'
-  schedule joins flow-match in `gen_worker.model.scheduler`: `EulerDiscrete` + `DiscreteSchedule`,
-  closed over the family's own declared block, importing neither a model library nor `torch`.
-  Differenced against `diffusers.EulerDiscreteScheduler` over 66 configurations — leading/trailing/
-  linspace spacing x epsilon/v-prediction (paired with `rescale_betas_zero_snr`, as
-  `gen_worker.view` pairs them) x every step count the sdxl and sd15 endpoints reach — and the
-  sigmas, the timesteps and `init_noise_sigma` are **0 ULP**, with a 28-step loop of
-  `scale_model_input` + `step` equal element for element. `IMPLEMENTED` is now TOTAL over
-  `SchedulerKind`, so `Sdxl` gains a typed `scheduler()`.
+- **`euler_discrete` is bare typed math, and it is MORE reproducible than the object it replaces.**
+  The U-Net families' schedule joins flow-match in `gen_worker.model.scheduler`: `EulerDiscrete` +
+  `DiscreteSchedule`, closed over the family's own declared block, importing neither a model
+  library nor `torch`. `IMPLEMENTED` is now TOTAL over `SchedulerKind`, so `Sdxl` gains a typed
+  `scheduler()`.
 
-  The bar was one ULP and hitting zero took reproducing the reference's PRECISION, not just its
-  algebra: its ladder is built by torch in float32 and interpolated by numpy in float64, and the
-  honest float64 reading is 201 float32 ULP off on the trained table (25 on a 28-step ladder)
-  because `alphas_cumprod` is a thousand-term cumulative product. Three narrowings are reproduced
-  exactly — `torch.linspace` float32 walks outward from both ends with a float32 step (the naive
-  form disagrees on 307 of 1000 entries), `torch.cumprod` float32 accumulates in **double** and
-  narrows only on store, and `1-ac` and the quotient are separate float32 ops. One more with a
-  wider blast radius: a Python float handed to a tensor operator arrives at its full double value,
-  so every scalar crossing that boundary is narrowed first — computing `sample / (sigma**2+1)**0.5`
-  in double makes every element of the result differ. Karras/exponential/beta sigmas are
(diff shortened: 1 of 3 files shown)


PaulFidika:
## Update: 8 ULP was also wrong. The INSTRUMENT was wrong, not the number.

CI refused `<= 8` with values up to **85**, after refusing `== 0` with values up to 2. Two wrong
bounds in a row is the signal that the quantity is not boundable — so this push changes the
measure rather than the constant, and that is the end of the iteration.

### Why no ULP bound can work

Every bit-level disagreement traces to a torch primitive that is **implementation-defined rather
than IEEE-exact**. Three, each surfaced by one CI cycle:

| primitive | behaviour |
|---|---|
| `torch.linspace` f32 CPU | dispatches by ISA — scalar vs vectorized differ on **145/1000** entries by 1 ULP |
| `torch.cumprod` f32 CPU | varies in accumulator width and association order — **dominant** term over a 1000-term product |
| `x ** 0.5` on a tensor | `pow`, **not correctly rounded**, where this module uses `math.sqrt`, which is |

The third only surfaced after I injected our ladder into the reference and the loop *still* differed
on CI — which also proves the first two are the table's problem, not the step's.

**diffusers cannot be bit-matched by anything, including diffusers.**

### What is asserted now

- **sigmas / `init_noise_sigma`** — relative, `<= 2e-4`. That is ~40× the observed 85-ULP (5.1e-6) cross-machine spread, and ~20× **tighter** than one bf16 ULP (3.9e-3), the precision the denoiser actually computes in.
- **timesteps** — still **exact**. Integer arithmetic; never differed on any machine, so 1 ULP there would be a real defect rather than library noise.
- **the loop** — relative in the **L2 norm**, not element-wise. Latents cross zero, so an element-wise relative measure divides by the noise floor and reports **17%** for a parts-per-million difference. Measured — that is precisely what the previous draft would have reported.

### Two measurements worth keeping

1. **With identical ladders the loop is bit-exact here** — all four configurations, both local CPU kernels. The step arithmetic is right; the residual was entirely the table.
2. **The loop propagates a ladder perturbation at gain ≈ 1.0** (5e-6 in → 4.9e-6 out). Well conditioned, no chaotic amplification.

Measurement 2 is why our cross-kernel byte-stability is a **correctness** property rather than a
nicety: two pods running diffusers can resolve measurably different ladders — and therefore
different latents — from one seed. Two pods running this module cannot.

Verified on `58cb4505`: **88 passed under auto-dispatch AND 88 under `ATEN_CPU_CAPABILITY=default`**;
`mypy src/gen_worker tests tests_v2` Success on 867 files; ruff clean.

PR #917, opened 2026-08-17
pgw#1355: a cold boot states its own decomposition — boot_stages, spans that overlap honestly

PaulFidika:
Closes pgw#1355. Paul's directive: *"every worker should be reporting metrics back to the hub, including cold boot time, broken down into stages (that should be overlapping ideally)."*

## What this is

e2e#1892 measured an 877 s cold serve and had to assemble the breakdown **by hand**, from four different event kinds:

| | source |
|---|---|
| 41 s queue | request-row timestamps |
| ~18 s boot/imports | `worker_boot_phases` (its own hub table, its own route) |
| 1.1 s pull, 2.5 s load | ditto |
| **804 s keyset derive** | `boot_adopt.duration_ms` on `worker_activity_events` |
| 6.2 s compute, 2.4 s finalize | the request row again |

Every one of those numbers already existed. What did not exist was a place where the SHAPE of a cold boot is stated once. So this is **a JOIN made durable at the source, not a new clock**.

`gen_worker.boot_stages` emits a versioned `boot_stages` series when the worker becomes ready:

- `phase=stage:<name>` — one row per span, `duration_ms` = its wall, `detail` = `v=1 stage= t0_ms= t1_ms= label= …`
- `phase=ready` — ONE terminal roll-up, `duration_ms` = **wall-to-ready**, `detail` packs the whole table so reading a pod's cold start is a one-row query

`k=v` detail in exactly the grammar the th#1839 route already serves verbatim (`(\w+)=(\S+)` parses it entirely, which e2e's `detailKV` already implements). **No hub schema change, no proto change, no migration.**

## The rule this refuses to break: the total is a UNION

Stages genuinely overlap on a real pod — a snapshot pull overlapping a component load overlapping a derive — so spans carry `t0_ms`/`t1_ms` offsets from OS process start and concurrency is visible rather than inferred.

```
span_sum_ms      = the sum of every span         (>= union whenever anything overlapped)
critical_path_ms = the UNION — wall in which SOMETHING ran
overlap_ms       = span_sum_ms - critical_path_ms
unmeasured_ms    = wall_ms - critical_path_ms    (NAMED, never smeared)
```

`boot_phases` learned this the expensive way; its own header records a summing reconciliation that "explained" 3,338 ms of a 909 ms fetch. A sum that exceeds wall is not a rounding problem — it is a report that has stopped describing time.

## The closed vocabulary

`process_boot`, `imports`, `hub_handshake`, `snapshot_pull`, `model_load`, `keyset`, `adopt`, `arm`, `warmup`, `ready` — an enum, because a renderer in another repo (e2e#1901) binds to these tokens. An unknown name raises at the emitter AND at the parser. A boot phase added to `boot_phases` with no stage and no documented exemption fails `unmapped_phases()`.

## The 805 s that had no span

pgw#1353 measured `worker_boot_phases` going `first_request_servable` then **871 s of silence** to `eager_ready`: the key-set derive runs during a REQUEST, after the boot window closed, so `boot_phases.in_boot()` is already False and no ladder row can cover it. `boot_adopt` now records it as one `keyset` stage carrying `keys_from` — the axis separating an 805 s boot from a 5 ms one — recorded **once per derive** rather than once per class, because 36 identical spans would report a boot whose stages sum to 36x its own wall.

The report is emitted at **`eager_ready`**, the first instant the worker could actually answer a request — deliberately NOT at `first_request_servable`, which on an eager-first boot fires before the model is loaded at all. That gap IS pgw#1353's 871 seconds.

## Red proofs (each verified to go red)

| break | result |
|---|---|
| `critical_path_ms` returns the SUM instead of the union | **5 tests fail** |
| the stage vocabulary stops refusing unknown names (emitter + parser) | **2 tests fail** |
| the over-long-span clamp stops confessing `clamped_ms` | **1 test fails** |

Plus a genuine finding the suite caught during development: `record_ending_now` was silently *shortening* a span whose duration exceeded process uptime. It now clamps (a negative offset is not representable) and states `clamped_ms`, because a silently shortened span is a wrong measurement that reads as a right one.

## Tests

`tests/test_boot_stages_pgw1355.py`, 25 tests, all green. Integration-style: the ladder tests drive the REAL `boot_phases` recorder with genuinely concurrent component fetches, and the emission tests drain the REAL activity sink the transport binds — so the assertions read exactly the ActivityUpdates a hub would receive.

mypy clean on every changed module.

## Not in this PR

- The renderer — e2e#1901.
- The hub-side twin decomposing `queue_ms` — th#2122.

### DIFF
--- a/changelog.d/pgw1355.md
+++ b/changelog.d/pgw1355.md
@@ -1,0 +1,54 @@
+- **pgw#1355: a cold boot reports its own decomposition, as one row.** e2e#1892
+  measured an 877 s cold serve and had to assemble the breakdown BY HAND out of
+  four different event kinds — request-row timestamps for the queue and the
+  compute, `worker_boot_phases` for the pull and the load, `boot_adopt`'s
+  `duration_ms` on `worker_activity_events` for the 805 s key-set derive. Every
+  number already existed; nowhere stated the SHAPE. `gen_worker.boot_stages`
+  emits a versioned `boot_stages` event series when a worker becomes ready: one
+  `phase=stage:<name>` row per span carrying `t0_ms`/`t1_ms` offsets from OS
+  process start, then one `phase=ready` roll-up whose `duration_ms` is
+  wall-to-ready and whose `detail` packs the whole table, so reading a pod's
+  cold start is a ONE-ROW query. `k=v` detail in the grammar the th#1839 route
+  already serves verbatim — **no hub schema change, no proto change, no
+  migration.**
+
+- **pgw#1355: overlap is first-class, and the total is a UNION.** Stages run
+  concurrently on a real pod — a snapshot pull overlapping a component load
+  overlapping a derive — so the report carries intervals rather than durations
+  and refuses to serialize what actually ran in parallel. `critical_path_ms` is
+  the union of the spans, `overlap_ms` the difference between the sum and the
+  union, and `unmeasured_ms` the wall no stage explained, NAMED rather than
+  smeared across the stages that were measured. A fixture with a pull inside a
+  load asserts the stages do not add up past wall; making the total a sum turns
+  five tests red. This is `boot_phases`' own lesson (its summing reconciliation
+  once "explained" 3,338 ms of a 909 ms fetch) applied at the reporting layer.
+
+- **pgw#1355: the stage vocabulary is CLOSED, and the reader says so when it
+  does not understand a fleet.** Ten stages — `process_boot`, `imports`,
+  `hub_handshake`, `snapshot_pull`, `model_load`, `keyset`, `adopt`, `arm`,
+  `warmup`, `ready` — as an enum, because a renderer in another repository binds
+  to these tokens and a stage invented at a call site is a column that appears
+  in production and in no reader. An unknown name raises at the emitter AND at
+  the parser; a boot phase added to `boot_phases` with no stage and no
+  documented exemption fails `unmapped_phases()`.
+
+- **pgw#1355: the derive that dominates every cold sdxl boot finally has a
+  span.** pgw#1353 measured `worker_boot_phases` going `first_request_servable`
+  and then **871 s of silence** to `eager_ready`, because the key-set derive runs
+  during a REQUEST — after the boot window closed, so `boot_phases.in_boot()` is
+  already False and no ladder row can cover it. `boot_adopt` now records it as
+  one `keyset` stage carrying `keys_from`, recorded once per derive rather than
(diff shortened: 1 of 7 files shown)


PaulFidika:
## Follow-up pushed: the report can never cost more than the boot it reports on

`boot_phases` buffers up to **2048** rows. Emitting one activity event per span unconditionally therefore had a worst case of a ~2000-message burst on the worker→hub stream **at exactly the moment the pod is trying to start serving** — the report costing more than the thing it reports on.

Capped at `MAX_STAGE_ROWS = 64`, **longest span first**. The ordering is the load-bearing half: dropping an arbitrary tail would be as likely to discard an 805 s derive as a 2 ms key fold, and the expensive stage being visible is the entire point of the event.

The **roll-up is never the row that gets dropped** — it is emitted after the children and carries the packed table, so a truncated series still answers the question the event exists for. `rows_truncated=<n>` says so, rather than leaving a reader to infer it from a stage having fewer rows than it had spans.

Two new tests (27 total, all green), each with a verified red proof:

| break | result |
|---|---|
| emit every span with no cap | `test_the_report_can_never_cost_more_than_the_boot_it_reports_on` **fails** |
| truncate in arbitrary order instead of longest-first | `test_truncation_keeps_the_spans_worth_reading` **fails** |

Also re-ran the adjacent suites against this branch — `test_boot_phases_pgw764`, `test_boot_span_ladder_pgw797`, `test_cold_boot_decomposition_pgw1087`, `test_activity_gw601`, `test_boot_adopt_observability_pgw1116`: **67 passed, 2 skipped**, no regressions.

---

### $0 live verification of this PR's central claim

Read against the standing `master` stack (`:31550`, hub `b10e245a57`). No pod rented.

```
GET /v1/admin/worker-activity-events?kind=boot_stages&limit=5   ->  HTTP 200  {"events": 0}
```

**A kind with no producer anywhere in the fleet is served with a 200 and an empty set.** That is this PR's whole premise — the th#1839 route stores every kind generically and serves `detail` verbatim — confirmed against the running hub rather than argued from the Go source. It also makes the absent-instrument path the *correct* reading of production today, which is what e2e#1901's renderer must print.

The control read, same hub, same minute — the premise this issue was filed on, still live:

```json
{"kind":"boot_adopt","phase":"miss","duration_ms":807695,
 "detail":"family=sdxl function=generate key=cg-key-v1-f92d1f9f763abc1e19ee88efa9"}
```

**807.7 s**, right now, and the only reason anyone can see it is that somebody went looking. That is the number this PR makes self-reporting.

PR #918, opened 2026-08-17
pgw#1354: a jobs-only image boots, is CUDA-probed, and a boot that finds nothing says so on the wire

PaulFidika:
Closes pgw#1354 (P0). Root-caused live by the e2e#1890 lane on two card families; fixed and red/green-verified here at $0 (no pods).

## What was broken

`entrypoint.get_modules_from_manifest` walked `manifest["functions"]` and nothing else. te#218 re-authored every conversion package as **jobs-only** (`conversion` 0.12.4 = 27 jobs / 0 functions), and those declare their modules in `jobs[].module` — which `discovery/discover.py` writes for every job and nobody read. So the first pod ever booted off a jobs-only image discovered ZERO user modules and hit `if not user_modules: return 1` **before hello**:

```
pod jfv904ztg8g7cm  RTX PRO 6000 Blackwell  created 20:37:10Z  dead 20:38:02Z
  worker_fatal: phase=compute_crash_loop deaths=2 deaths_before_hello=2 last_cause=exit:1
  [hardware-unsuitable] driver="" gpu=""     ready_at = NULL
```

pgw#1324's `worker.py` fix (`if not specs and not jobs`) is correct and sits **one frame too late** — `collect_jobs([])` over an empty module list finds nothing either. Its tests construct the executor directly and never traverse the entrypoint's module extraction, which is how a green suite shipped a wheel that cannot boot the one package shape the jobs program exists to run.

## Three changes

**1. One walk, both blocks.** New `gen_worker.manifest_blocks` owns the declaration-row derivation over `functions[]` + `jobs[]`; `get_modules_from_manifest` reads it. Deliberately not a second jobs walk beside the function walk — two walks drift, and the next block makes it three (the th#2105 precedent: one derivation, two feeders). Stdlib-only, no `gen_worker` imports, because the control parent reads it and must stay a bare interpreter (pgw#763).

**2. The CUDA probe stops being blind to jobs.** `should_probe_cuda` read `functions[]` too, so a jobs-only GPU image returned `False` and was **never health-probed** — gw#529's bad-host guard silently did not apply to the whole jobs program, and 10 of those 27 conversion jobs ask for a card. It now reads the same derivation. The mixed CPU/GPU rule (defer to the installed torch build) is unchanged and now applies *across* blocks rather than per block, which is the right answer: a release publishing separate `accelerator=none`/`accelerator=cuda` images is one release either way.

**3. The silent death dies too.** The no-modules exit was the **one** fatal in `_run_main` that did not dial — a bare `logger.error` + `return 1`. RunPod exposes no container-logs API, so the hub saw `exit:1` with no reason class and classified it `[hardware-unsuitable]` with empty driver/gpu: a boot bug wearing a hardware verdict, which is exactly why this took a paid pod to find. It now dials `_log_worker_fatal("no_user_modules", …, settings=settings)` over the worker_fatal carrier (pre-Hello reporting was deliberately preserved in th#2075), with a detail that NAMES the gap — how many functions and jobs the lock declared, how many modules that yielded, which blocks this build walks — and discriminates "declarations this wheel cannot read" from "no baked manifest at all", because those have different owners. `manifest_loaded` gained `job_count` beside `function_count` for the same reason.

## The class sweep (item 4)

This is the **fourth** instance of teach-the-next-layer (ingest th#2100 → lane derivation th#2105 → pick floor th#2120 → entrypoint here). Every `functions[]` walk in `src/`, classified:

| site | verdict |
|---|---|
| `entrypoint.get_modules_from_manifest` | **the defect** — now both blocks |
| `cuda_probe.should_probe_cuda` | **instance #5, found by this sweep** — now both blocks |
| `entrypoint` `manifest_loaded` `function_count` | **blind telemetry** — a boot log that read "0 declarations" on exactly the images the jobs program runs. `job_count` added |
| `discovery/validation.validate_endpoint_lock` — the `len(functions)==0` warning | **cosmetically wrong** — called a 27-job release "will advertise nothing". Now jobs-aware |
| `discovery/validation` — slot-layout, wire-route, aot-precondition walks | **correctly functions-only**: slots, bindings, compile cells and wire routes are concepts a job does not declare (`_job_entry`: "no execution lanes, no compile block, no slots/bindings") |
| `discovery/discover.main` refusal | **already jobs-aware** (`not functions and not jobs`) |
| `models/download.build_provider_index_from_manifest` | **correctly functions-only**: it indexes `functions[].bindings`, and a job that wants a hub-resolved model NAMES it in its payload rather than binding a slot |
| `cli/serve.py`, `cli/listing.py` | **not a manifest walk** — local-dev CLI over in-process registry candidates, no `endpoint.lock` read |

## Red/green, at $0

`tests/test_jobs_only_boot_pgw1354.py` — 9 arms, all driving production code. The packages are real `@job`/`@endpoint` sources; the `endpoint.lock` is baked by the **real** `python -m gen_worker.discovery` in a real subprocess (the generated Dockerfile's own build step); the boot is the real `_run_main` through the real `Worker.__init__`/registry walk. Only the CUDA device, the cache-dir mkdir and `Worker.run`'s gRPC loop are stood down. The jobs-only fixture is the measured shape: 27 jobs across 2 modules, 10 on a card, zero functions.

Four mutations, each re-run against the full file:

| mutation | red |
|---|---|
| restore the `functions`-only loop in `get_modules_from_manifest` (= master) | 3 failed — including the boot at `code == 1`, no Worker ever constructed, which is `deaths_before_hello=2` for $0 |
| restore `functions = manifest.get("functions")` in `should_probe_cuda` | 2 failed |
| delete the `_log_worker_fatal` call from the no-modules exit | 2 failed |
| keep the call but drop `settings=` (the dial silently becomes a no-op) | 2 failed |

Green: `9 passed`. Regression sweep over every file touching these modules — `313 passed`.

## Sequencing

Every jobs-only endpoint is dead-on-boot until a wheel ships this. jobs#230 must land before the jobs repo repins (its Dockerfile smoke-imports).

### DIFF
--- a/changelog.d/pgw1354.md
+++ b/changelog.d/pgw1354.md
@@ -1,0 +1,23 @@
+- **A JOBS-ONLY IMAGE BOOTS.** `entrypoint.get_modules_from_manifest` walked `manifest["functions"]`
+  and nothing else, so a package that declares its modules in `jobs[].module` — which is every
+  package te#218 re-authored, e.g. `conversion` 0.12.4 at 27 jobs / 0 functions — discovered ZERO
+  user modules and exited 1 before hello. Measured on rented hardware twice, on two card families
+  (e2e#1890): `deaths_before_hello=2`, `ready_at` NULL, no pod has ever reached hello off a
+  jobs-only image. pgw#1324's `worker.py` fix (`if not specs and not jobs`) is correct and sat one
+  frame too late — `collect_jobs([])` over an empty module list finds nothing either. The walk now
+  reads BOTH declaration blocks, from one derivation (`gen_worker.manifest_blocks`) rather than a
+  second walk beside the first.
+- **A jobs-only GPU image is CUDA-probed.** `cuda_probe.should_probe_cuda` read `functions[]` too,
+  so gw#529's bad-host guard silently did not apply to the jobs program: 10 of those 27 conversion
+  jobs ask for a card, on a device nothing had health-checked. It now reads the same both-block
+  derivation; the mixed CPU/GPU rule (defer to the installed torch build) is unchanged and applies
+  across blocks rather than per block.
+- **The no-modules exit dials a typed fatal.** It was the one fatal in `_run_main` that did not —
+  a bare `logger.error` + `return 1`. RunPod exposes no container-logs API, so the hub saw
+  `last_cause=exit:1` with no reason class and condemned the pod `[hardware-unsuitable]` with empty
+  driver/gpu: a boot bug wearing a hardware verdict, which is why this took a paid pod to find. It
+  now reports `no_user_modules` over the worker_fatal carrier with a detail that NAMES the gap —
+  how many functions and jobs the lock declared, how many modules that yielded, and which blocks
+  this build walks — and discriminates "declarations this wheel cannot read" from "no baked
+  manifest at all". The boot log's `manifest_loaded` phase gained `job_count` beside
+  `function_count` for the same reason.
--- a/src/gen_worker/cuda_probe.py
+++ b/src/gen_worker/cuda_probe.py
@@ -14,6 +14,7 @@ from typing import Any, Optional
 
 import msgspec
 from .hostfacts import cuda_ready
+from .manifest_blocks import declaration_rows
 
 CUDA_PROBE_FAILED_MARKER = "GEN_WORKER_CUDA_PROBE_FAILED"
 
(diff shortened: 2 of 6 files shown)


PaulFidika:
**CI note.** `fast gates` (the master ruleset's only required context) is GREEN in 2m18s: mypy strict over `src/gen_worker tests tests_v2`, all 30-odd tree guards, `uv build`. Every one also run locally first.

`tests` is RED with 31 failures, and every one of them is in `tests/test_sd_stage1_pgw1346.py` — another lane's euler_discrete ULP comparison, which this branch does not touch. **Master is red on the same file, worse**: run 32066691149 on `a8e94316` reports 59 failed / 5183 passed, all at `test_sd_stage1_pgw1346.py:132/168/174/214`. The file arrived with `ff9e1657` (pgw#1346 B2), which is master's commit, not this branch's, and a sibling lane holds the fix (`1346-b2-ulp-fix`).

Per `docs/releasing.md`: *"If the red is a semantics decision inside another lane's work, do not fix it to unblock your tag."* Not touched here, not laundered here.

PR #920, opened 2026-08-17
pgw#1356: a mint that finished says so, and names the classes it produced

PaulFidika:
## The measurement

A rented A40 (2026-08-17, pod `zco8e1bx0t1jgk`, $0.54) ran a mint to completion for the first time: two compile children, **36 of 36 graph classes `status=compiled exit=0`**, pool efficiency 99.89 %. The one roll-up the hub sees read

```
aot_mint_phases in_flight status=in_flight n_entries=0
  totals={'export_s': 0, 'compile_s': 0, 'warm_s': 0} phases={} overlays={}
```

Sixty minutes of measured work, reported as *"still running, nothing produced"* — indistinguishable from a mint the pod killed, which is exactly the distinction pgw#848 built the on-disk snapshot to preserve.

## Two independent defects

**1. The reader and the writer disagreed about an address.** `mint_graph_classes` stamps the phase table onto `MintedArtifact.mint_phases`; `mint_supervisor._mint_phase_table_of` read `entry.metadata["mint_phases"]`. `metadata` is TCG's **closed** artifact vocabulary and cannot legally carry the field — `test_tcg_mint_parent_pgw1270` asserts `"mint_phases" not in survivor.metadata` by name. So the reader could never find it, `phase_table` fell through to the on-disk snapshot on *every successful mint*, and the roll-up inherited the terminus a snapshot stamps: `in_flight`, because a snapshot is a beat.

**2. The table's entry rows have had no writer since pgw#1215.** `_mint_phase_table` folded `MintProgress.minted`, a list the SERIAL driver appended to as it exported in-process. pgw#1215 deleted that driver (`aot_mint.mint` no longer exists) and left the field with no writer — so `n_entries` was `0`, `entries`/`phases`/`overlays` were empty for a mint of any size, and every per-class `entry:<name>` row was suppressed downstream by `if entry_s <= 0: continue`.

The table is now folded out of `EntryCompilePool.class_spans`, per GRAPH CLASS, over the **survivor** set so an absorbed duplicate is not double-counted. `phases` folds the flat `torchcg.spans` partition labels the child actually writes; `overlays` come from the pool's per-SHARE ledger, which is the granularity they are measured at — the child discards its per-class overlay split (`_overlays`), and inventing a per-class attribution would be worse than reporting none.

`MintProgress` **binds** the pool's live dicts rather than copying them, so the snapshot rewritten on every beat names the classes that have landed. pgw#848's own worked example — attempt sixteen, killed at class 30 of 36, reporting zero rows — has been the real behaviour of every mint since the serial driver died.

## Proof

`tests/test_mint_rollup_terminus_pgw1356.py` — **6/6 red on `origin/master`, 6/6 green here**, driven over the REAL relay (`mint_graph_classes` → `_mint_phase_table_of` → `phase_table` → `_emit_aot_phases`), because the defect is a seam between those four and any one of them tested alone reads green. Only the spawn of compile children is replaced (a real AOTI compile is a pod leg). The red run reproduces the production row character-for-character:

```
assert 'n_entries=3' in "family=sdxl lane=bf16-w16a16 status=in_flight n_entries=0
  totals={'export_s': 0, 'compile_s': 0, 'warm_s': 0, 'total_s': 0.0} phases={} overlays={}"
```

Blast radius **500 passed**; the 8 failures are the torch-missing baseline `origin/master` shows identically (same 8 names). mypy delta **zero** over 869 files. All 16 hard lint gates pass.

## What this is NOT

Not the `key_axis_divergence` refusal the same A40 mint also hit. **That one is already fixed on `master` by pgw#1340 (PR #895) and is UNRELEASED**: tag `v0.122.0` is `3b10ede1`, cut 3h28m *before* #895 merged, and `serverless-endpoints/sdxl` pins `gen-worker==0.122.0`. The wheel the pod ran still compared `family` at a seam no TCG artifact can state it across. **That half needs a release cut, not a code change.**

Closes the telemetry half of pgw#1356.

### DIFF
--- a/changelog.d/pgw1356.md
+++ b/changelog.d/pgw1356.md
@@ -1,0 +1,32 @@
+- **A mint that FINISHED reported itself as still in flight, having produced nothing.** Measured
+  on a rented A40 (2026-08-17, pod `zco8e1bx0t1jgk`, $0.54): two compile children, 36 of 36 graph
+  classes `status=compiled exit=0`, pool efficiency 99.89 % — and the one roll-up the hub sees read
+  `aot_mint_phases in_flight status=in_flight n_entries=0 totals={'export_s': 0, 'compile_s': 0}
+  phases={} overlays={}`. An operator could not tell a finished mint from one the pod killed, which
+  is precisely the distinction pgw#848 built the on-disk snapshot to preserve. Two independent
+  defects, both required for the row to be readable: `mint_supervisor._mint_phase_table_of` read
+  the table from `entry.metadata["mint_phases"]` while the writer stamps `MintedArtifact
+  .mint_phases` — and `metadata` is TCG's CLOSED artifact vocabulary, which cannot legally carry
+  the field (`test_tcg_mint_parent_pgw1270` asserts its absence by name), so the reader could never
+  succeed and `phase_table` fell through to the snapshot on EVERY successful mint, inheriting the
+  snapshot's `in_flight` terminus. The reader now reads the typed field.
+- **The phase table's entry rows have had no writer since pgw#1215.** `_mint_phase_table` folded
+  `MintProgress.minted` — a list the SERIAL driver appended to as it exported in-process. pgw#1215
+  deleted that driver (`aot_mint.mint` no longer exists) and left the field with no writer, so
+  `n_entries` was `0`, and `entries`, `phases` and `overlays` were empty, for a mint of any size:
+  every per-class `entry:<name>` row the hub should have received was suppressed by an
+  `if not timings: continue`. The table is now folded out of `EntryCompilePool.class_spans`, per
+  GRAPH CLASS, straight off the children's reports, over the SURVIVOR set so an absorbed duplicate
+  is not double-counted. `phases` folds the flat `torchcg.spans` partition labels the child
+  actually writes; `overlays` come from the pool's per-SHARE ledger, which is the granularity they
+  are measured at — the child discards its per-class overlay split, and inventing a per-class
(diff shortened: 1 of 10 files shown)


PaulFidika:
**Update — swept the remaining call sites, rebased over #914.**

The first CI run went red on two files a fully green local blast radius could not see: `tests/test_pool_sizing_pgw842.py` and `tests/test_child_traces_its_own_share_pgw1215.py` **skip entirely on the dev box** (`could not import 'torch'` — the `--extra dev` install fails on a `torch==2.13.0+cu130` hash mismatch), so a signature change they call is first observed on a hosted runner.

Fixed by **grep sweep, not by re-running** — a re-run of a skipped file proves nothing:

- `_mint_phase_table([], …)` → `({}, …)` in `test_pool_sizing_pgw842.py` (×2) and `test_mint_oom_classification_pgw848.py` (also torch-gated; its pre-existing red would have hidden a new one).
- `class_spans`/`entry_overlays` on the `_Pool` double in `test_child_traces_its_own_share_pgw1215.py`, for the same reason the pgw#1270 and pgw#824 doubles needed them.

Verified zero remaining: no `_mint_phase_table([` anywhere, and the only monkeypatched `EntryCompilePool` doubles in the tree are the four now carrying both attributes. Every other `EntryCompilePool` reference constructs a real pool.

Rebased onto `origin/master` after #914 landed (21:24Z) and cleared the `test_sd_stage1_pgw1346.py` red, so this run is measured against a green base rather than against master's known breakage.

PR #921, opened 2026-08-17
pgw#1354: cut 0.123.0 — a jobs-only image can boot

PaulFidika:
**Release cut: 0.123.0.** Pinned at `cf82ae4d` (master's tip at cut time), 77 commits since `v0.122.0`, 17 fragments assembled.

## Why cut now, against rule 1

`docs/releasing.md` rule 1 is *"cut when a POD RUN needs one, not when a lane finishes something"* — and a pod run needs one, in the strongest form the rule admits:

- **Every jobs-only endpoint is dead on boot until a wheel carries pgw#1354.** `conversion` 0.12.4 is 27 jobs / 0 functions, and no pod has ever reached hello from a jobs-only image — measured twice, on two card families, at `deaths_before_hello=2` with `ready_at` NULL. The e2e#1890 lane has already paid for that evidence.
- **th#2120, the other blocker it measured, is merged hub-side** (`507ed2f7`). This wheel is the last thing standing between the jobs program and a live pod run.
- **pgw#1340** is a running cost, not a latent one: every self-mint since 0.117.0 buys its 20-45 minute compile and then refuses itself, and *"every pod on 0.117.0-0.122.0 keeps paying"* until a wheel ships the fix.

Batched, per the same rule: everything ready at this moment rides.

## The sweep for silent release notes

`git ls-tree origin/master changelog.d/` against `git log v0.122.0..origin/master --merges` found **three merges that filed no fragment** — pgw#1344 (the HF→GGUF tensor plane through tensorfs), pgw#1350 (a named sibling that is not on disk is a REFUSAL), pgw#1352 (the sm-clearance probe, plus the per-pod bring-up budget and the re-roll's host-not-budget rule). Written here by the cutter, from the commit messages, because a merge with no fragment is a silent release note and lanes forget.

`assemble_changelog.py` also pruned the seven 0.121.0-consumed fragments and recorded all 17 rows in `consumed.tsv`. Nothing was deleted by hand.

## Range verified BY ANCESTRY, not from cut notes

`git rev-list --count v0.122.0..HEAD` = **77**. `git merge-base --is-ancestor <c> HEAD` confirmed for each must-ride: `cf82ae4d`/`3bb1df94` (pgw#1354), `83f1c454` (pgw#1340), `adb00d1f` (pgw#1341), `b0bebc2e` (pgw#1306), `43f55534` (pgw#1328), `706c4c60` (pgw#1352), `7f08a63d` (pgw#1344), `9b44b359` (pgw#1350).

## Publish proof

Dispatched CI on this exact tree — run [32072116899](https://github.com/cozy-creator/python-gen-worker/actions/runs/32072116899), `workflow_dispatch`, not a `pull_request` run. `publish.yaml` refuses `only_pull_request_proof`, so this is the evidence the tag will key on.

**The tag goes on this branch commit, so this PR must land as a TRUE MERGE** — which the queue's configured method already is. A squash would rewrite the commit, orphan the tag, and make the next cutter's `git rev-list v0.123.0..HEAD` silently re-list everything released here.

## Downstream, owed but NOT this PR

- **jobs#230 must land before the `jobs` repo repins** — pgw#1306 deletes `ConversionContext`, and three Dockerfile smoke-imports break the IMAGE BUILD, not a test.
- **pgw#1357** (all 27 endpoint families pin `gen-worker==`, zero ranges) is why this wheel still needs 27 file edits to reach a pod. That is the sweep that makes the next cut cheap.


PaulFidika:
**From the pgw#1356 lane — two facts this cut needs, one good, one that will be misread if not stated.**

### ✅ This cut lands a P0. Please don't let it slip.

`0.123.0` consumes `pgw1340`, and I verified the code is really on this branch (`git merge-base --is-ancestor 83f1c454 b6356a58` → true). That is the fix for **pgw#1356 / th#2098**: every self-mint on the fleet currently compiles for 20–60 minutes and then refuses its own artifact with `key_axis_divergence — family: child cell states '', this runtime computed 'sdxl'`, because the arm seam compared three axes TCG's closed artifact vocabulary has no field for. Measured cost: a 36/36-class A40 mint discarded at $0.54, `compiled_graph_store` still at 0 rows fleet-wide, and e2e#1892's journey steps 2+4 blocked on it. **This wheel is the unblock.**

### ⚠️ The tagging step is where the last one went wrong — please check it explicitly

`v0.122.0` is tagged at `3b10ede1`, the **branch-tip bump commit**, not the merge commit. pgw#1340 merged at 14:28:23Z; that tag was cut at 11:00:17Z. `git merge-base --is-ancestor 83f1c454 3b10ede1` → **false**. So a fix that was on `master` for three and a half hours was absent from the wheel every pod pulled, and nobody noticed until an A40 reproduced it eight hours later. Suggested guard, cheap and one line per claim:

```
git merge-base --is-ancestor <fix-sha> <tag-sha>   # for anything the CHANGELOG says ships
```

### ⚠️ pgw#1356 is NOT in this cut, and its absence produces a false negative

PR #920 merged at **21:53:18Z** (`73147932`), after this branch was created — `changelog.d/pgw1356.md` is not here and the code is not on `b6356a58`. That is fine to defer, **but whoever re-runs the mint on 0.123.0 must be told**, because the defect it fixes is a lying success signal:

> On 0.123.0 a mint that compiles all 36 classes and **publishes correctly** will still emit
> `aot_mint_phases status=in_flight n_entries=0 phases={} overlays={}`.

The roll-up reader looked for the phase table at an address TCG's closed metadata can never hold, so it fell through to the on-disk beat snapshot — whose terminus is `in_flight` — and `MintProgress.minted` has had no writer since pgw#1215, so `n_entries` is `0` for a mint of any size. **Judge 0.123.0's mint by `compiled_graph_store` rows and the `self_mint_publish` legs, not by the `aot_mint_phases` row.** The `status=minted n_entries=36` signal only becomes trustworthy in the wheel after this one.

Happy to be told this belongs in a follow-up cut instead — flagging it here only because #921 is in the queue now.

PR #928, opened 2026-08-17
pgw#1346 B4: Wan 2.2 is THREE models, LTX needs no new scheduler math, and F2 dissolves

PaulFidika:
The video DiT batch of [[#1346]] W2, **authoring only** (no endpoint migration — that half is
gated on W1b-3; no pods, no cards, no weights). Five declarations land, and three of the batch
plan's own rows are refuted by measurement rather than executed.

## What lands

| model | tier | runners | buckets | scheduler |
|---|---|---|---|---|
| `Wan22T2vA14b` | graph | `denoiser_high`, `denoiser_low` | `shape` x `frames` = 2x1 | **none** (K10) |
| `Wan22I2vA14b` | graph | `denoiser_high`, `denoiser_low` | 2x1 | **none** (K10) |
| `Wan22Ti2v5b` | graph | `denoiser` | 1x1 | **none** (K10) |
| `Ltx23` | graph | `denoiser` | `video_tokens` x `audio_tokens` = 2x2 | `flow_match_euler_discrete` |
| `MINIMAX_H3` | eager | — | — | — (F2) |

Nine fake-tensor exports, four committed export/binding pairs, 30 new tests.

## Three verdicts the plan did not have

**1. Wan is THREE models, not the plan's `Wan22Ti2v` + `Wan22A14b`.** On B1's rule (class =
architecture config, instance = weights + tuned + ref), measured against the published
`transformer/config.json` of all three checkpoints:

| key | T2V-A14B | I2V-A14B | TI2V-5B |
|---|---|---|---|
| `in_channels` | 16 | **36** | **48** |
| `out_channels` | 16 | 16 | **48** |
| `num_layers` | 40 | 40 | **30** |
| `num_attention_heads` | 40 | 40 | **24** |
| `ffn_dim` | 13824 | 13824 | **14336** |

I2V's 36 is not a width tweak — the endpoint's own mint file spells it `"channel_concat_36 =
16 noisy + 4 mask + 16 cond"` against T2V's `"text_only"`, so it is a different first
convolution and a different traced call. TI2V additionally takes a **per-token float32**
timestep where the A14B pair takes one **scalar per batch**, i.e. the calls differ in rank and
torchcg G2 makes one runner one signature. The hub agrees: `wan-22-t2v-a14b`, `wan-22-i2v-a14b`
and `wan-22-ti2v-5b` are three separately registered strings, and the repo ships three
`aot/*.mint.json`. Asserted at the export, not at the source: the three resolve three different
ingresses from the same bucket coordinate.

**A14B really is two DiTs, and the loop is the better place to say so.** The expert pair
publishes BYTE-IDENTICAL configs, so it is ONE graph class run twice — two runners over two
counted stages, `steps_high` then `steps_low`. That states the MoE budget that diffusers leaves
implicit in a `boundary_ratio` threshold, which is the endpoint's own complaint about upstream:
*"the split moved silently whenever the shift moved."*

**2. F2 dissolves — both horns rest on premises this lane measured to be false.**

F2 asked whether H3's bespoke loop projects onto a `Scheduler` block or needs `loop: host`, and
offered "eager `Model` (losing the compiled backing it has today)" versus "the `Scheduler`
vocabulary grows a host-implemented escape".

* **There is no minted backing to lose.** minimax-h3 is the only video endpoint with **no `aot/`
  directory and no mint declaration**, and its own source says why: *"THIS FAMILY IS NOT
  AOT-DECLARABLE (ie#652: 737k static classes)."* What it runs is `torch.compile` over one
  dynamic sequence axis, which the eager tier does not touch.
* **The `Scheduler` block does not need a host escape because it already IS one.** recipe_v1:
  *"torchcg validates its shape and never interprets it: the host implements the named
  scheduler."* `MiniMaxH3Scheduler`'s entire config surface is one positive float.
* **The loop is STAGED, and `host` would be the lie.** Per step the body is `["denoiser",
  "update"]` in fixed order, of which only the first is a graph (`update` is the scheduler step,
  host by definition); video and audio are ONE fused forward returning two tensors, not two
  alternating runners. Neither of H3's loops is data-dependent — the denoise runs a payload
  `Literal[20, 30, 50]` and `long_video` runs `1..MAX_SLOTS` with `MAX_SLOTS = 24` declared on
  the wire. `loop.kind: host` REFUSES a repeat count, so declaring it would drop a bound H3
  genuinely has. There is a red proof of exactly that in the suite.

**So H3 is eager for three reasons, none of which is the loop or the scheduler**, each recorded
on the declaration: (a) no architecture source in this repo — the transformer, both VAEs and the
scheduler are vendored in the endpoint at a pinned diffusers SHA and diffusers 0.39 ships no
MiniMax class, so `Runner.build` has nothing to construct (same shape as B1's `Flux2Klein9b`
finding); (b) the scheduler PARAMETERS are stamped tuned values (`video_shift`, `audio_shift`),
and G11 makes the class-level document structurally checkpoint-free; (c) te#171's AdaLN cache is
keyed by the RESOLVED SCHEDULE — `CacheKey(steps, video_shift, audio_shift)` — with a forward
pre-hook rewriting `timestep`/`timestep_indices` every step, so a baked denoiser's identity would
depend on a request's step count.

**3. The scheduler math owed by B4 is one item, not three.**

* **"wan needs a distilled flow-match" — CLOSED by measurement.** `FlowMatchEulerDiscrete` under
  a static shift already resolves `sigma_i = s*x/(1+(s-1)*x)` over `x_i = (N-i)/N`, which is term
  for term wan's own `distilled_sigmas` fed through its `shifted_sigma`. Asserted against both
  live-verified ladders: 4 steps at shift 5.0 gives `[1000, 937.5, 833.3, 625]`, 8 steps gives all
  eight. The endpoint had to subclass diffusers only because diffusers double-shifts (a 4-step
  shift-5 run lands on t=24 instead of 625); this module never had that bug.
* **"ltx's literal ladders — hard dep on B3's explicit-sigma work" — NO DEPENDENCY.** `Schedule`
  is already a public frozen dataclass over an explicit sigma tuple; synthesis from a step count
  is what `FlowMatchEulerDiscrete` adds on top. `schedule_from_sigmas` is three lines and refuses
  a stamped ladder that carries the terminal zero.
* **UniPC-on-flow-sigmas is NOT owed either** — B3-math's #923 landed it mid-lane and this branch
  rebased onto it. **So B4's scheduler-math column is EMPTY, and that is what makes K10 the whole
  blocker**: both of Wan's solvers are implemented and the three declarations still cannot carry
  them, because `GraphModelSpec.scheduler` is ONE block and codegen emits ONE `scheduler()`. The
  obstacle was never a missing kind. The declaration keeps the served 27-key UniPC config quoted
  as the reference for whoever wires K10's scheduler set.

## K10 recurs, and Wan is the stronger case

All three Wan models declare **no scheduler block**. One model serves TWO solvers selected by
which adapter a request attaches — the checkpoint's own `UniPCMultistepScheduler`
(`use_flow_sigmas=True`, `prediction_type="flow_prediction"`, `solver_order=2`, `bh2`) for the
base lineage, flow-match Euler on the trained uniform ladder for the distilled one. **No payload
enum is involved**, unlike B2's sdxl/sd15 case: this is the checkpoint's own trained solver.
A single-valued block would name one and silently serve the other lane a schedule it was not
trained on, so all three name none — `GraphModelSpec.scheduler` is optional exactly so a family
with no single honest answer can decline, and codegen then emits no `scheduler()`, which makes
the miss an `AttributeError` on the author's machine.

## K9-video, disposed in two halves

* **Wan needs nothing beyond K9.** `frames` is a genuine product axis (one trained temporal grid
  per family), so `shape` x `frames` is 2x1 on A14B and 1x1 on TI2V — **exactly** the shipped mint
  rows, no phantom cell. The packed `(w,h)` trick is owed the same fix it was owed in B2, for the
  same two numbers.
* **LTX is where the product genuinely fails — and not on a spatial axis.** Its graph is a
  function of token counts alone (`num_frames`/`height`/`width`/`fps` are dead inside the forward
  because `video_coords` is precomputed, so 704x1280 and 1280x704 are ONE graph), and the real
  axes are `T_v` x `T_a` x `T_at` x `T_kf`: **82 graph classes on h100 / 115 on b200** over
  **28 / 39** distinct `(T_v, T_a)` pairs, roughly 3x smaller than the axis product. Packing them
  would make a 9-digit literal hiding a sparse set behind a dense-looking axis. **Filed as
  K9-video, not contorted.** What IS declared is the sub-product that is exact: the 4K/241-frame
  lane's stage-1 and stage-2 token counts x the two audio lengths 241 frames can carry, four
  variants, two of which are the committed mint rows byte for byte.

## NEW BLOCKER K11 — `ModelSpec.name` conflates two identities

`name=` is both the generated symbol root (torchcg G1 constrains it to `[a-z][a-z0-9_]*`, no
hyphens) and the `register_family` key tensorhub validates repo metadata against. Every hub
architecture string these four families carry is hyphenated, and tensorhub's `Normalize` drops
only `.` and spaces — so the two spellings cannot meet. **Wan makes it structural rather than
cosmetic: three GRAPH identities share ONE stamped vocabulary root (`wan22`)**, so even a
hyphen-tolerant name would not fix it. The clean shape is a `tuned_root=` separate from `name=`.
Filed as a red proof rather than prose (B1 has the same gap on `flux2_klein_4b`).

## ie#740 floors, migrated BY VALUE

Asserted as PARSED NUMBERS, not strings: T2V `sm89+` + `vram80g`; I2V `vram80g` **and no fp8
lane** (deliberate — *"fp8 was never measured on the I2V lane"*, so uniformising would claim a
rung nobody measured); TI2V `vram24g` (the whole point of per-lane floors: a 5B model that sat
behind the A14B pair's 80); LTX `sm89+` + `vram78g` (*"two B200s at $6.83/hr rented and
refused"*); H3 `vram78g` (th#1754) — on the EAGER tier, which proves the layout axes are the
model's and survive a model having no graph classes.

## Incidental fix

Codegen no longer imports `MappingProxyType` into a scheduler-less binding (its only use is
freezing `SCHEDULER_PARAMETERS`). The three Wan models are the catalog's first scheduler-less
families, so nothing had surfaced it. **No existing export digest moves** — the four pre-existing
generated modules regenerate byte-identically.

## Gates

- `mypy src/gen_worker tests/test_video_dit_pgw1346.py` — **Success, 370 files**
- `tests/test_video_dit_pgw1346.py` — **30 passed**; three mutations verified RED (I2V
  `in_channels` 36->16, `vram80g`->`vram79g`, a dropped LTX sigma)
- neighbouring SDK suites `test_flux_stage1_pgw1331` + `test_model_sdk_pgw1332` +
  `test_sdk_v2_pgw647` + `test_sd_stage1_pgw1346` + `test_export_contract_pgw739` — **300 passed**
- serve-role and codegen suites `test_adopt_only_serve_role_pgw1328` + `test_pad32_codegen_pgw1079`
  + both guard-closure suites + `test_torchless_boot_pgw788` + `test_reset_family_registry_pgw1161`
  — **73 passed**
- `scripts/check_model_bindings.py` — **8 pair(s) current**
- `scripts/lint_serve_role_closure.py` — green, 49 declared roots (up from 39)
- every other `fast gates` lint green; `ruff check src/gen_worker` clean
- **No GPU rented, no weights downloaded, no network.** Every export is fake-tensor; the loop
  proofs run on `.fake()` backings.

## What B4 does NOT do

The endpoint migration half (3 dirs, 8 `@endpoint`s, 19 `ctx.defaults` reads, 3 pins) is gated on
W1b-3, exactly as B1 and B2 are. K5's siblings (RIFE `interpolator` in all three dirs, wan's
`AnimeUpscalerPipeline` `finisher`, ltx's `LTX2LatentUpsamplePipeline` `upsampler`) are still
unowned: their classes live in the ENDPOINT repos, so an eager `ModelSpec` for them can carry
`layouts=` by value but has no `build=`. F1's recommendation is right in shape and needs a home
for the class.

**Three rebases pulled three mid-lane landings into this PR** — B3b's #925, B3-math's #923 with
B5's #922, and W1b-2's #916 — and each one changed a real answer rather than only the base: #923
emptied the scheduler column, #922 made `MINIMAX_H3` a type, and #916 supplied
`Runner(component=)`, which is the field the A14B expert pair was missing. See the comments.


PaulFidika:
Rebased onto `origin/master` @ `1371fe96` (B3b's #925 landed mid-lane). Three index files conflicted and were resolved by UNION, which is what they are: `catalog/__init__.py`, `serve/role.py` and `check_model_bindings.py` all just gained rows. B3b's two eager models (`anima`, `hidream_o1`) and B4's `minimax_h3` are held by the same real walk for the same reason — the model's own code is not reachable from this SDK, so none of them takes a guarded-import hatch.

Re-verified on the rebased head:

- `mypy src/gen_worker tests tests_v2` (the CI command verbatim) — **Success, 897 files**
- `scripts/check_model_bindings.py` — **10 pair(s) current** (was 8 before B3b, 4 before B1)
- `scripts/lint_serve_role_closure.py` — green: **58 declared roots**, 49 of them model-free, 7 enumerated guarded edges
- `scripts/assemble_changelog.py --check` — green
- `ruff check src/gen_worker` — clean
- targeted suites, **338 passed**: `test_video_dit_pgw1346` (30) + `test_flux_stage1_pgw1331` + `test_model_sdk_pgw1332` + `test_sdk_v2_pgw647` + `test_sd_stage1_pgw1346` + `test_export_contract_pgw739` + `test_adopt_only_serve_role_pgw1328` + `test_pad32_codegen_pgw1079`
- the neighbouring catalog suites B1/B3b landed, **111 passed**: `test_flux2_klein_4b` + `test_flux2_klein_9b` + `test_anima` + `test_hidream_o1`

No export digest moved in the rebase.

PaulFidika:
Rebased again onto `origin/master` @ `b2070af4`, which brought in **two things that directly change B4's owed column**:

**1. `unipc_multistep` LANDED (B3-math, #923) — and K10 gets sharper, not softer.** B4's original row said UniPC was the one scheduler math still owed. It is not owed any more: it is in `SchedulerKind` and `IMPLEMENTED` today, verified by B3-math at all three served `flow_shift` values (12.0 / 5.0 / 3.0). Combined with B4's own proof that the distilled lane's ladder IS `flow_match_euler_discrete` exactly, **both of Wan's solvers are now implemented and the three declarations still cannot carry them.** The obstacle was never a missing kind — `GraphModelSpec.scheduler` is ONE block and codegen emits ONE `scheduler()`, so a family whose CHECKPOINT chooses the solver has nowhere to put the second one. `scheduler=None` stays, and there is a new test asserting exactly this (`test_k10_is_now_a_declaration_limit_and_not_a_missing_implementation`). **K10's own fix — a scheduler SET keyed by tuned name — is the only thing these three are waiting on.**

**2. B5's eager-tier codegen LANDED (#922), so `MINIMAX_H3` is now a TYPE.** `MinimaxH3` joins the catalog with a committed `minimax_h3.eager.json` + binding: a real `Model` subclass carrying the tuned schema and nothing else — no bucket literals, no runner callables, no `scheduler()`, and `variant()` refuses. That closes the one thing F2's eager verdict was still missing (something an endpoint can annotate a handler parameter with) **without** claiming a composition H3 does not have.

Re-verified on the new head:

- `mypy src/gen_worker tests tests_v2` — **Success, 921 files**
- `scripts/check_model_bindings.py` — **10 graph + 12 eager pair(s) current**
- `scripts/lint_serve_role_closure.py` — green: **74 declared roots**, 65 model-free
- every other `fast gates` lint green; `ruff check src/gen_worker` clean; changelog gate green
- **451 passed**: this suite (32, up from 30) + the five neighbouring SDK suites + the serve-role and codegen suites + all four catalog suites B1/B3b landed
- **316 passed** on the two suites the rebase pulled in and B4 interacts with: `test_dit_solvers_pgw1346` (B3-math) and `test_b5_eager_models_pgw1346` (B5)

The three conflicting files are the same additive index files as before (`catalog/__init__.py`, `serve/role.py`, `check_model_bindings.py`) — resolved by union, with the duplicate import rows deduped and the index verified consistent (46 exported names, `_FAMILIES` and `__all__` agree exactly, no duplicates).

PaulFidika:
Rebased a third time, onto `origin/master` @ `bcd17263` — **W1b-2's injection path (#916) landed, and it brings the field these declarations were missing.**

`Runner(component=)` now exists, and A14B is the family W1b-2's own finding ("no runner→component map exists anywhere in the repo") was missing it FOR: two runners that share ONE graph class are distinguishable only by where they live in the loaded tree. So the Wan declarations name them —

| model | runner | component |
|---|---|---|
| `Wan22T2vA14b` / `Wan22I2vA14b` | `denoiser_high` | `transformer` |
| `Wan22T2vA14b` / `Wan22I2vA14b` | `denoiser_low` | `transformer_2` |
| `Wan22Ti2v5b` | `denoiser` | `transformer` (dense — no `transformer_2`) |
| `Ltx23` | `denoiser` | `transformer` (the endpoint's own `Compile(targets=("transformer",))`) |

— which turns "the expert pair is two weight sets over one graph class" from a comment into the module an eager backing actually reaches.

**`component` is a SERVING fact and does not ride the export digest**, asserted two ways rather than assumed: `check_model_bindings.py` is still byte-clean at all 22 pairs with the committed documents UNCHANGED, and a new test decodes `wan22_t2v_a14b.export.json` and asserts the string `transformer_2` does not appear in it at all. Restating a serving fact must not re-key an artifact — the same claim B2 made for the ie#740 floors, on the other new field.

Re-verified on this head:

- `mypy src/gen_worker tests tests_v2` — **Success, 923 files**
- `scripts/check_model_bindings.py` — **10 graph + 12 eager pair(s) current**
- all 18 `fast gates` lints green (incl. `lint_serve_role_closure.py`: 74 declared roots, 65 model-free); `ruff` clean; changelog gate green
- **704 passed** across this suite (34, up from 32) + `test_model_injection_pgw1346` (W1b-2's own suite, since this PR now uses its field) + the five neighbouring SDK suites + serve-role/codegen + B3-math's `test_dit_solvers_pgw1346` + B5's `test_b5_eager_models_pgw1346` + all four B1/B3b catalog suites

Only `changelog.d/pgw1346.md` conflicted this time (the shared fragment ~10 lanes edit); the index files merged clean.

PaulFidika:
## ⚠️ `tests` came back RED — diagnosed, and it is NOT this PR. Recording it before the queue does anything, because pgw#911's censure is exactly this situation.

**`1 failed, 5708 passed, 38 skipped`.** The one failure is
`tests/test_procsplit_pgw763.py::test_parent_exit_with_unretired_results_is_reported_typed`:

```
assert h.wait_exit(60.0) == 0
assert h.pc.unretired_results_at_exit == 1, captured_dials
E   AssertionError: []
E   assert 0 == 1
WARNING gen_worker.transport: connection to 127.0.0.1:41763 failed: ConnectionResetError: Connection lost
stream error StatusCode.UNAVAILABLE: ... 127.0.0.1:45377: Connection refused
```

**Four independent reasons it is not mine, stated as checks rather than as a claim:**

1. **Zero coupling.** The complete non-catalog surface of this PR is two files: `model/codegen.py` (one conditional import line, reachable only from `gen-worker model generate`) and `serve/role.py` (tuples of module-name strings). The procsplit suite references neither, and contains no catalog reference at all.
2. **It passes locally on this exact branch** — `tests/test_procsplit_pgw763.py` **20 passed** in 168s.
3. **The failure signature is a race, not a logic error.** The test spawns a fake child that writes a durable `JobResult` and exits 0 *before the parent has any stream* (its own docstring), then asserts the parent booked exactly one unretired result. It got zero — the parent tore down before the child's write landed. Both log lines are connection-teardown, not assertion-of-values.
4. **The test is timing-fragile by construction:** `stop_flush_timeout_s=2.0` is a hard 2-second flush window used as a correctness gate. On a 4-core `ubuntu-latest` running the suite under xdist, a 2s window is a coin flip. This is the "procsplit starvation flake" B2's lane already recorded in the tracker (*"the only local failure was an unrelated procsplit starvation flake"*), and it is the no-magic-timeouts shape: a wall-clock budget standing in for progress toward a goal.

**I have rerun the failed job rather than asserting flakiness** — evidence, not an argument. Result posted below when it lands.

**Filed as owed, because it is a real defect in the TEST and not in the worker:** `stop_flush_timeout_s=2.0` should be a progress condition (wait until the child's result is observable or the child is reaped), not an elapsed-time budget. Whoever owns pgw#763 should take it; this lane is not touching an unrelated suite to go green.

Required context `fast gates` is **green** on this head, and the lane's own targeted set is **704 passed** locally on this exact tree, including all 34 of this PR's tests.

PaulFidika:
## ✅ Rerun of the SAME commit: `tests` **pass** (13m9s). The flake is now demonstrated, not argued.

`fast gates` pass · `tests` pass — both green on head `2e881c06`, with no code change between the red run and the green one. That is the controlled experiment: same tree, same job, opposite outcome, which is what "flake" means and what the previous run alone could not establish.

The diagnosis above stands and gets one more confirmation worth recording: `SplitHarness` defaults to `stop_flush_timeout_s=30.0`, and `test_procsplit_pgw763.py:701` is the **only call site in the suite** that lowers it to `2.0`. So the fragility is local to that single test rather than to the worker's flush path — which is both why it is safe to attribute away from this PR, and why the fix belongs to pgw#763 rather than here.

**Still owed (not taken in this lane, deliberately):** that 2s window should gate on a progress condition — the child's durable result is observable, or the child has been reaped — never on elapsed wall-clock. Recorded in the tracker under B4's status block.

PaulFidika:
Rebased onto `origin/master` @ `e0d8b6fe` (B3a's families #929 and its fragment split #932). Head is now `57a5098f`.

**Adopted B3a's per-lane changelog convention, which is the recorded fix for the thing that keeps re-serializing this queue.** B4's notes moved out of the shared `changelog.d/pgw1346.md` into **`changelog.d/pgw1346-b4-video.md`**; `pgw1346.md` is now byte-identical to master's copy, so this PR no longer contends on the file ~10 lanes edit. That file was the ONLY conflict in three of my four rebases.

Also corrected one now-stale sentence in the fragment: it still said UniPC was "the only scheduler math B4 leaves owed". It is not owed — B3-math landed it. **B4's scheduler-math column is empty, and that is what makes K10 the whole blocker.**

One thing worth flagging because a union-merge hides it: resolving four rebases of three additive index files by union quietly introduced **9 duplicate rows** (6 in `catalog._FAMILIES`/`__all__`, 3 in `role.MODEL_FREE_MODULES`). Ruff caught the dict half (`F601`); the tuple half is invisible to every gate because a duplicated string in a membership tuple changes no behaviour. Both deduped and asserted rather than eyeballed: `_FAMILIES` and `__all__` are 59 entries with an empty symmetric difference, `MODEL_FREE_MODULES` is 76 with no repeats, and all five B4 symbols resolve through the lazy `__getattr__`.

Verified on `57a5098f`:

- `mypy src/gen_worker tests tests_v2` — **Success, 946 files**
- `scripts/check_model_bindings.py` — **14 graph + 12 eager pair(s) current**
- `scripts/lint_serve_role_closure.py` — green: 84 declared roots, 75 model-free, 11 guarded edges
- all 17 remaining `fast gates` lints green; ruff clean; changelog gate green
- **593 passed** on the SDK/injection/solver/eager set, **167 passed** on every catalog suite in the repo (B1, B3a, B3b, B4)

PR #933, opened 2026-08-17
pgw#1353 option (b): the key set outlives the ENDPOINT, not just the pod — `keys_from=hub`

PaulFidika:
The worker half of **pgw#1353 option (b)**. Hub half: tensorhub th#2123, PR cozy-creator/tensorhub#1406, **merged `e18940df`**.

## What was still broken after row (a)

Row (a) (PR #915) put the `cg-keyset-v1` document on th#1813's durable root and fixed every endpoint that **has** one. An **ephemeral private deployment** has none — no network volume, no baked document, a fresh container per pod — so it still paid the measured **778-833 s** of `torch.export` once per pod, forever. That is 94 % of `worker_queue_ms` on a cold sdxl serve, on four independent pods.

## The tier

```
shipped → durable → this pod's cache → THE HUB → derive
```

`src/gen_worker/keyset/hub.py` — `GET`/`PUT /v1/worker/keysets/<closure_digest>` through the existing parent-mediated broker, with `HubTier` carrying *who to ask* as one value instead of two strings threaded through five call sites. A hub hit reports `KeySource.HUB` → **`keys_from=hub`** on the `boot_adopt` event: its own member, for the same reason `DURABLE` is one. The four tiers answer four different operational questions — *baked* / *this endpoint has durable storage* / *the platform holds it for an endpoint that has neither* / *somebody paid 805 s* — and collapsing any pair makes "which fix is reaching the fleet" unanswerable off the boot events.

**Admission is the address check** (§4.29, as amended by Paul's keys-shipped-as-data ruling). The answer must NAME the closure it was asked at, or it is refused and the pod derives. That is what makes it safe for an untrusted pod to have written the row: a writer can only make a statement about a closure it could already derive itself, and a lie costs a resolve MISS plus a re-mint — never a wrong graph.

**Asked last among the readers, before the deriver.** A local document is already on this machine, so a round trip in front of it is pure added boot latency; and a 10 s bound against 805 s of tracing needs no argument.

## The producer

On a derive, the document is offered to the hub — **best effort, never blocking**. It runs strictly after the keys this boot needs already exist, so no failure of it can slow, block or change this boot. A pod that traced for 805 s and could not upload has still booted, and the reason is logged.

The row is **read back from the writable root the deriver just wrote**, not rebuilt from `DerivedKeySet`: that type carries *folded* keys with this pod's `sm` and toolchain baked in, and a document built from those would be a machine-dependent answer stored at a machine-independent address — wrong for every pod on another SKU.

There is deliberately **no "publish only TRACED" guard**. This runs on the derive branch, where every read tier *including the hub* has already missed, so a `boot_key.derive` memo hit is exactly as worth publishing as a fresh trace. That guard reads plausible and would strand every cozy-local memo hit; it is mutation-tested as an **added** guard for exactly that reason.

`scripts/emit_cg_keyset.py --publish` fills the store from the **mint lane**, which is what makes *"a serving pod never derives"* true rather than *"a serving pod derives once per closure"*. Refused BY NAME without `--hub`/`$COZY_HUB_URL` and `$COZY_WORKER_TOKEN`; a publish that did not land exits non-zero.

## The asymmetry, stated because it is a design choice

A malformed document on a **local** root is still `keyset_invalid` and still propagates — a mint-lane or image defect that must be visible as itself. A **hub** answer that does not parse, or names another closure, is a MISS with a recorded reason. The hub is a network peer on the boot path; a pod that refuses to boot because a cache was unreachable is strictly worse than the cost it was avoiding.

## ⚠️ One other lane's fence is narrowed

`test_a_populated_store_and_an_unreachable_hub_makes_zero_http_attempts` (pgw#1127) fenced §4.28's *"never requested"* **at the socket**. §4.28 is about the **artifact**, and that property is **unweakened** — it is now asserted DIRECTLY (`resolve_batch` is never called) instead of through a socket proxy, plus the only address dialled must be the key-set route.

The argument, written into the test rather than left in a commit message: that pod holds no key-set document (its deriver is a patched lambda, which hides that the real one would trace), so the honest description of it is *"a pod about to spend 13 minutes"*, and one 10 s bounded ask with a fallthrough is strictly better for it. **A pod genuinely told there is no hub still dials nothing at all** — that is the test above it, and it still fences at the socket.

## Proof

`tests/test_keyset_hub_tier_pgw1353b.py`: two `boot_adopt.attempt` calls with disjoint caches, **no durable root** (the ephemeral shape), one in-memory hub between them that **reimplements th#2123's contract** rather than mocking this client's calls — a mock that returned whatever the client asked for would prove the client talks to itself. Pod 2 boots with an armed `torch.export` sentinel, an armed `subprocess.Popen` sentinel, and a deriver that fails the test if reached.

**Mutation audit — 11 severed or added guards, `red 5/5` each:**

| mutation | verdict |
|---|---|
| the hub read deleted from `key_set_from_data` | **red 5/5** |
| the publish deleted from `_key_set` | **red 5/5** |
| a hub hit reports `keys_from=memo` | **red 5/5** |
| a plausible "publish only TRACED" guard *added* | **red 5/5** |
| the read tier not distinguished (hub-read pod publishes back) | **red 5/5** |
| admission accepts an answer at the wrong address | **red 5/5** |
| a hub fault propagates instead of degrading | **red 5/5** |
| a failed upload propagates instead of degrading | **red 5/5** |
| the producer read-back guard narrowed past `OSError` | **red 5/5** |
| the hub asked BEFORE the local roots | **red 5/5** |
| a corrupt LOCAL document degrades to a miss | **red 5/5** |

Blast radius **186 tests** green across the keyset, boot-adopt, cell-resolve and boot-key files. mypy clean over 937 files. All 24 hard lint gates and `assemble_changelog --check` pass.

## Wheel

`changelog.d/pgw1361.md` — **rides 0.125.0**, the next cut. `0.124.0` already carries row (a). The fleet needs to do nothing until the next repin: the hub tier is read by new wheels only, and a hub with the routes and a wheel without them is exactly today's behaviour.

### DIFF
--- a/changelog.d/pgw1361.md
+++ b/changelog.d/pgw1361.md
@@ -31,3 +31,26 @@
   closure"*. Refused BY NAME without `--hub`/`$COZY_HUB_URL` and `$COZY_WORKER_TOKEN`, and a
   publish that did not land is a non-zero exit — a `--publish` that quietly did nothing is
   exactly the failure this issue is about.
+
+- **TWO SEAM DEFECTS FOUND ON THE WAY, and either alone would have made the tier DEAD ON EVERY
+  FLEET POD — silently.** Under the process split the compute child holds no credential and
+  never names a free-form path: the parent refuses any request not matching
+  `procsplit/actions.py` by method AND full-path regex.
+
+  1. **`keysets.fetch` / `keysets.publish` were not in the action table.** Every unit test that
+     patches `broker.request` stays green with the tier completely unreachable — the parent's
+     refusal arrives as a `BrokerError`, `keyset.hub` correctly reads it as a miss, and the pod
+     derives for 805 s exactly as before. That is pgw#1309's shape precisely, so it is now
+     asserted against the TABLE rather than against the client. The path regex pins the address
+     grammar to the 32 lowercase hex `parse_closure_digest` admits; a looser `[^/]+` would let a
+     process running tenant code put anything it liked in a path segment the parent attaches the
+     pod's credential to. `keysets.publish` is deliberately NOT in `PUBLISH_ACTIONS`: pgw#980
+     disarms those on a live-edit probe because a probe writing a CELL poisons every later
+     adopter, and a key set carries no artifact and lands at a closure only that probe resolves.
+
+  2. **`parent._http_call` attached the body only when `method == "POST"`** — true while POST was
+     the only verb in the table, and silently fatal the moment a PUT joined it. A key-set publish
+     would have gone out with NO BODY, the hub would have answered a typed `keyset_invalid`, and
+     the child would have read a well-formed "no" to a request it believed it sent. Now every
+     body-bearing verb (POST/PUT/PATCH) sends what the action table already authorized; GET and
+     DELETE still send none, which is what they declare.
--- a/changelog.d/pgw1361.md
+++ b/changelog.d/pgw1361.md
@@ -1,0 +1,33 @@
+- **`keyset.store` gains a HUB tier, so a pod with no volume and no baked document stops paying
+  the 805 s.** Search order is now **shipped → durable → this pod's cache → the HUB → derive**,
+  and a hub hit reports `keys_from=hub` — its own `KeySource` member, so "which fix is reaching
+  the fleet" stays a query rather than a story.
+
+  pgw#1327 shipped the `cg-keyset-v1` document as a file; pgw#1353 row (a) put it on th#1813's
+  platform-placed durable root. Neither reaches an **ephemeral private deployment** — no network
+  volume, no baked document, a fresh container per pod — which measured **778-833 s of
+  `torch.export` per cold boot on four independent sdxl pods**, 94 % of `worker_queue_ms`, once
+  per pod forever. th#2123's `GET`/`PUT /v1/worker/keysets/<closure_digest>` is the one store
+  every pod shape can reach.
+
+  **On a successful derive the document is offered to the hub, best effort**, so a fleet
+  converges instead of paying per pod. The publish runs strictly AFTER the keys this boot needs
+  already exist, and **nothing it can do reaches the boot**: a hub that is down, a 409 from
+  another pod's document, a volume that vanished mid-read — every one is a logged sentence and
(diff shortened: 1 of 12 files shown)


PaulFidika:
## ⚠️ Correction to this PR's own claims — pushed `da2b7109`

The head this PR opened at was **green and dead on every fleet pod**, and the reason is worth reading because it is not specific to this change.

Under the process split — the real fleet configuration — the compute child holds no credential and **never names a free-form path**: the parent refuses any request that does not match `procsplit/actions.py` by method AND full-path regex. Two things were wrong there:

1. **`keysets.fetch` / `keysets.publish` were not in the action table.** The parent's refusal arrives as a `BrokerError`; `keyset.hub` correctly reads it as a miss (it must never block a boot); the pod derives for 805 s exactly as it does today. **Every test in the lane stayed green**, because every one of them patches `broker.request` — which is the *child's* side of the seam. That is pgw#1309's shape precisely.

2. **`parent._http_call` attached the body only on `method == "POST"`.** True while POST was the only verb in the table, and silently fatal the moment a PUT joined it: the publish would go out with **no body**, the hub would answer a typed `keyset_invalid`, and the child would read a well-formed "no" to a request it believed it sent.

Both are fixed, and the fix is asserted **against the action table**, not against the client. Four new guards, mutation-audited **red 5/5** each — the actions dropped from the allowlist; the address grammar loosened to `[^/]+`; the parent dropping the body on any verb but POST; a key-set publish swept into `PUBLISH_ACTIONS`. **Lane total: 15/15 red 5/5.**

Two design notes that came out of it:

- the path regex pins the address grammar to the **32 lowercase hex** `parse_closure_digest` admits, not `[^/]+` — a process running tenant code must not be able to put anything it likes into a path segment the parent attaches the pod's credential to;
- `keysets.publish` is deliberately **not** in `PUBLISH_ACTIONS`. pgw#980 disarms those on a live-edit probe because a probe writing a *cell* into a shared family namespace poisons every pod that later adopts from it. A key set carries no artifact and no signature and lands at a closure only that probe resolves, so sweeping it in would disarm the tier on exactly the pods a dev loop uses most.

The PR description above is otherwise accurate; the mutation table there is now the 11-row subset of a 15-row audit.

PR #936, opened 2026-08-18
pgw#1362 wave 2: the cli/config/sdk cluster folds from 9 incident files into 3 domain modules

PaulFidika:
Second wave of the pgw#1362 fold. **Stacked logically on [#935](https://github.com/cozy-creator/python-gen-worker/pull/935) but branched from `master`**, so the two are independent; whichever lands second rebases over a one-line `HIGH_WATER` and a handful of removed `pyproject.toml` list entries.

## The fold

**9 modules / 2,660 lines → 3 modules / 2,226 lines.** −67% files, −16% lines.

| new module | what it is |
|---|---|
| `test_config_authority.py` | the declared interpreter env, and the torch flags it imposes |
| `test_config_boundary.py` | who may read settings, and how a reconcile resolves |
| `test_sdk_authoring.py` | the decorators, the job contract, what a handler emits |

Plus `test_model_sdk_pgw1332.py` → **`test_model_sdk.py`**: 1,320 lines, already single-domain, already torch-gated. It needed a *name*, not a merge, and forcing it into one would have bought a worse file.

**All 120 tests survive** — counted on both trees before the originals were deleted.

## The torch trap, and why the split is what it is

A module-level `pytest.importorskip("torch")` **poisons a merge**: it converts a per-module skip into a silent skip of every neighbour that never needed torch. So `test_settings_authority_pgw1049` and `test_env_contract_pgw718` (both already gated, both torch in every test) merged **with each other**, and the three torchless config modules merged **separately**. This rule was written down in wave 1 and this is the first wave it bound.

## Strictness, again up not down

Four more modules come off the mypy `ignore_errors` burn-down and **none go back on** — high-water 255 → 251 on this branch, composing with wave 1's cut. `mypy src tests` green over 937 files; all 24 fast-gate lint scripts pass locally.

## Vacuity discipline

**One redundant autouse fixture deleted rather than carried:** pgw#718's `_reset_boot_seal` unwound `env_seal._BOOT_READBACK`, which pgw#1049's `_fresh_boot` — now module-wide — already unwinds along with `_ESTABLISHED_OVERRIDES`. Only reset-shaped autouse fixtures may widen on a merge (they can only add isolation); a strictly-subsumed one is dead weight.

One load-bearing guard per merged module was proved able to go RED against a real production mutation:

| module | mutation | result |
|---|---|---|
| `config_authority` | `settings_authority.impose_torch` stops imposing `cudnn_benchmark` | 2 tests RED |
| `config_boundary` | `runtime_config.observe` stops writing the snapshot | 2 tests RED |
| `sdk_authoring` | `api/jobs` stops refusing a class-declared `@job` | 1 test RED |

## Left alone on purpose

`test_import_cycles_pgw981`, `test_retired_env_names_pgw1305` and `test_cozy_runtime_env_contract_pgw1237` are **guards** (the last a corpus-digest gate). The epic's brief leaves the guard family alone; they are merged only within their own kind, if ever.

Three prose cross-references that named folded files moved with them: `src/gen_worker/api/jobs.py`, `scripts/lint_serving_process_compiles.py`, `tests/test_job_dispatch_pgw1324.py`.

## Verification

- `pytest tests/test_{config_authority,config_boundary,sdk_authoring}.py` → **120 passed**
- `mypy src tests` → **Success, 937 files**
- 24 fast-gate lint scripts, incl. `lint_mypy_ratchet.py` at the lowered mark → all pass
- `assemble_changelog.py --check` → OK
- Full `tests` suite runs here in CI. **Not enqueuing until it is green.**


PaulFidika:
**Ruling landed while this was in CI — recording the effect on this PR honestly.**

Paul, 2026-08-17, on the epic's target: *"yeah but can we delete a bunch of low-value tests we don't need? unfortunately your coding agents tend to keep adding more complexity, and never remove any of it. So we have an asymmetry where the tests balloon. Just because a test exists doesn't mean it's useful. You can probably do a few integration tests and get the same or better coverage."*

That makes **value**, not vacuity, the unit of judgment, and asks for **a few broad integration journeys** in place of clusters of narrow rigs. Coverage is now counted in *properties guarded*, and every wave reports four counts: deleted-as-vacuous / deleted-as-low-value / folded-into-journey / kept-focused.

**This PR — like #935 — is a PURE FOLD cut before that ruling: 9 modules into 3, 120 tests in and 120 out, nothing deleted.** It is correct and it lands (Paul confirmed #935 under the new ruling and said to continue by cluster), but it is the *floor* of what a wave should now do, not the model. Wave 3 (procsplit, 25 files / 8,487 lines) opens with a property inventory and a value pass, and only folds what survives — the plan is on pgw#1362.

No change requested here; this comment exists so the PR record does not read as if the fold were the whole job.

PR #947, opened 2026-08-18
pgw#1374: Paul's comment-purge mandate, third repo — 463 comment lines, and the measurement that says why it is not 8,828

PaulFidika:
Third repo of Paul's comment-purge mandate, after serverless-endpoints
([[ie#747]], 8,828 lines / 31.4%) and tensorhub ([[th#2136]], 6,015 / 9.7% of the
addressable set). Same classifier, same green bar.

## What landed

**463 own-line comment + docstring lines removed** over 70 files:

| bucket | before | after | removed | ratio |
|---|---|---|---|---|
| `src/gen_worker` (.py) | 53,797 | 53,451 | 346 | 0.64% |
| `.github/workflows` (.yaml) | 806 | 705 | **101** | **12.5%** |
| `scripts/` (.py) | 4,212 | 4,196 | 16 | 0.38% |
| **addressable total** | **60,121** | **59,658** | **463** | **0.77%** |

The class deleted is the one both predecessors named and nothing else: campaign
play-by-play (`224 rows on sd15 and 28 on ernie across wheels 0.117.0 and
0.118.0`), backstories for code that no longer exists (`this used to be
begin_fleet_mint, which additionally re-pointed…`), CI archaeology (a benchmark
table comparing a superseded single-job workflow shape against its replacement),
and measurement narratives whose durable half is one CITATION. Every design
record, kill-predicate, owner, fence-asserted marker and single-sentence WHY
stays.

## The finding, and it is the point of this PR

**pgw is not serverless-endpoints and it is not even tensorhub. The 80% Paul
estimated is not available here, and the number is 0.8%, not 31%.**

th#2136 built, ran and did NOT ship a blanket "condense every multi-paragraph
comment to its first paragraph" rule after review showed it deleting the KEEP
class. The same rule was measured here:

| | lines | share |
|---|---|---|
| addressable Python prose (`src/`+`scripts/`+`examples/`, minus vendored/generated/worktree-locked) | 50,604 | — |
| what the blanket first-paragraph rule would delete | **28,407** | **56.1%** |

Twelve paragraphs were sampled at random from the 5,520 that rule would drop.
**Twelve of twelve are the KEEP class** — a load-bearing `-inf` in
`dpm_multistep._log` ("the infinity is load-bearing, not an edge case to clamp
away"), the w8a8 snapshot verifier's three-way tolerance contract, the SDK
`Asset` stored-vs-inline description, `flux1_dev_serve`'s structural argument for
why a serve module may not import its declaration, the llama fit ladder's
degraded-not-dead contract. Not one is historical narrative.

Three independent strata were sampled so that is not a regex artifact:

| stratum | sampled | were deletable |
|---|---|---|
| paragraphs the historical/campaign net flagged (3,022 lines / 467 paras / 178 files) | 480 lines read | ~15% |
| random non-first paragraphs >= 3 lines | 22 | 2 |
| random single-paragraph comment blocks of 2-6 lines (3,003 blocks / 9,767 lines) | 26 | **0** |

**pgw's Python comment corpus is design record and API reference, written
dense.** The historical residue is real, this PR removes it, and it is ~360
lines — not ~20k. Reaching Paul's 80% here means deleting what he said to keep.

**The one place the predecessors' ratio DOES transfer is the workflows** —
`ci.yaml` was 566 comment lines of 812 (70% of the file), the same shape
th#2136 found in `fast-gates.yaml` (489 -> 184). That is the 12.5% row above.

## A repair the sweep surfaced

`ci.yaml`'s `pgw#1310 gate` header had two lines spliced into the middle of a
sentence:

```
# difference is a whole class of defect this workflow could not see: for
# Naming what WAS required is the whole point of the sentence below.
# retired-name: historical rationale, not a live reference.
# two published releases `gen-worker`'s metadata required `hashrepo`, a
```

The splice was **load-bearing, not sloppiness**:
`lint_retired_package_names._classified` accepts its marker only on the finding's
own line or the one IMMEDIATELY above, and `hashrepo` is on the following line.
Reordered so the prose reads and the marker still classifies — the fence's
`--selftest` and its real scan are both green.

## Skipped, deliberately

* **`tests/` and `tests_v2/` ENTIRELY** (32,651 prose lines). [[#1362]]'s cluster
  table marks every remaining flat cluster TODO, its folds rewrite these comments
  anyway, and it has already MEASURED the tree-wide docstring collapse at 9,718
  lines as its own mechanical PR. Sweeping there collides with a live epic.
* `src/gen_worker/_vendor/` (digest-fenced) and `src/gen_worker/pb/` (generated).
* **`pyproject.toml`** — ie#747's finding #2: a comment in a pyproject costs a
  `[project].version` bump and therefore a relock. 142 comment lines left alone.
* `CHANGELOG.md` + `changelog.d/` — read as DATA by
  `test_changelog_attribution_pgw1339` and `test_changelog_consumption_pgw1226`.
* `docs/dockerfile.md` + `examples/micro-diffusion/Dockerfile` —
  `test_dockerfile_doc_contract_pgw1016` matches their RAW BYTES, comments
  included, because the hub's publish validator does.
* The seven `src/` files sibling worktrees are editing (`executor.py`,
  `aot_mint.py`, `aot_compile_pool.py`, `aot_compile_child.py`,
  `model/{backing,residency,runtime}.py`) plus
  `scripts/lint_{incident_test_names,mypy_ratchet}.py`.

## The presence-assertion inventory (th#2136's precondition, built FIRST)

125 `assert "…" in <text>` sites across `tests/`; 187 files read production
source. Nearly all assert CODE shape, which comment deletion cannot touch. The
comment-sensitive residue, all left untouched:

* `ValidationError.__doc__` must keep `"do not retry"` (`api/errors.py:8`,
  asserted at `test_declaration_refusal_pgw1075.py:212`) — verified present.
* The escape-hatch comment markers, counted before and after over
  `src/`+`scripts/`+`examples/`: `noqa` 303->303 · `type: ignore` 68->68 ·
  `pragma:` 123->123 · `tcg-vocab:` 15->15 · `fence-symbol-exempt:` 4->4 ·
  `refused:` 40->40 · `oci-image:` 3->3 · `mixed-cas-hatch:` 3->3 ·
  `pickle-ban:` 1->1. **All identical.**
* No fence in this repo asserts a comment's PRESENCE. Deleting comments can only
  ever *satisfy* a banned-vocabulary fence (`lint_tcg_vocabulary`,
  `lint_retired_*`), never break one — th#2136's finding, confirmed here.

## Green bar

* **AST identity per changed `.py`**, against the merge base, with docstrings
  normalised SEPARATELY (`ast.dump` of the tree with docstring `Expr` nodes
  stripped). 68/68 identical; the only reported differences are docstring text,
  reported by symbol name.
* **Both workflows `yaml.safe_load` to data byte-identical** to the merge base,
  and every replaced/inserted line is asserted to be a comment line by the
  applier.
* `mypy src/gen_worker tests tests_v2` -> **Success, 918 source files** (unchanged).
* `ruff check` clean; the full `fast gates` lint set green **including every
  `--selftest` arm** and `lint_unreached_surface --siblings-selftest`.
* The 20 presence-assertion test modules run locally: **224 passed**.
* Structural guards: **0 new orphaned `#:` sphinx-field blocks** (the three that
  exist are byte-identical to the base), and list/table lines 264 -> 266 —
  nothing dropped.
* `git diff --diff-filter=D --name-only origin/master HEAD` is empty.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,96 +1,40 @@
 name: CI
 
-# pgw#952 — this workflow is split into two PARALLEL jobs, and runs the full
-# suite on every PR. pgw#977 (2026-08-05) narrowed the trigger back to `master`
-# alone, because `dev` was deleted when the three-tier flow retired — lanes now
-# PR straight to `master`, so `[master]` IS the every-lane trigger. Both jobs
-# are the master ruleset's required contexts (`fast gates`, `tests`).
-#
-# ─── THE GAP THIS CLOSES, AND WHY IT STAYS CLOSED ──────────────────────────────
-#
-# Until pgw#952 a lane PR in this repository could run NO CI AT ALL. Verified on
-# PR #466 (`950-legacy-hardcut`, 215+/178-): `statusCheckRollup` was `[]`. Not a
-# check that passed — no check ran. At the time the cause was `ci.yaml` naming
-# only `master` while lanes targeted `dev`; the branch names have since swapped
-# roles, and the invariant that matters is the one to preserve:
+# pgw#952 — this workflow is split into two PARALLEL jobs and runs the full
+# suite on every PR.
 #
 #   **THIS WORKFLOW MUST TRIGGER ON WHATEVER BRANCH LANES ACTUALLY PR INTO.**
 #
 # If that branch is ever renamed or a second one added, change the list here in
 # the same commit. A lane merging with an empty check rollup is the failure this
 # file exists to make impossible, and a reader cannot tell it apart from "CI is
-# broken". pgw#949 is the same failure already written down: the integration
-# branch sat RED on two gw#666 guard tests since PR #455 and nothing said so.
+# broken" — it has happened (pgw#952, `statusCheckRollup: []`; pgw#949, a branch
+# red on two guard tests with nothing saying so).
 #
 # ─── WHY THE FULL SUITE IS PRE-MERGE HERE, UNLIKE TENSORHUB ────────────────────
 #
-# An older header said "CI spend is metered (WORKSPACE-GIT-POLICY.md)". That is
-# true of tensorhub, which is PRIVATE — and it is why tensorhub's `verify` moved
-# to a post-merge `push: [master]` trigger (th#1621/th#1114, ~24 runner-hours/day
-# at lane volume). **This repository is PUBLIC**
-# (`gh repo view --json visibility` -> PUBLIC), and GitHub Actions on standard
-# runners is free and unmetered for public repositories. `ubuntu-latest` is a
-# standard runner. That constraint does not exist on this repo, so the suite
-# stays pre-merge where it can actually block a defect.
-#
-# What remains real is LATENCY, not spend. That is what the two jobs below are
-# for: a lane gets the cheap verdict in ~1m35s and does not wait 15m to learn
-# that mypy is unhappy.
+# Tensorhub's `verify` moved to a post-merge `push: [master]` trigger because it
+# is PRIVATE and metered (th#1621/th#1114, ~24 runner-hours/day at lane volume).
+# **This repository is PUBLIC**, and GitHub Actions on standard runners is free
+# and unmetered for public repositories — so the suite stays pre-merge where it
+# can actually block a defect. What remains real is LATENCY, not spend, which is
+# what the two jobs are for: the cheap verdict in ~1m40s instead of 15m.
 #
 # ─── MEASURED, ON THE RUNNER ───────────────────────────────────────────────────
 #
-# Not extrapolated from this 32-core box — a 4-core `ubuntu-latest` behaves
+# Never extrapolate from this 32-core box — a 4-core `ubuntu-latest` behaves
 # nothing like it, which is the trap th#1572's lane fell into by a factor of 40.
-# Left column: run 30847566437, the last green run of the old single-job shape,
-# re-attributed to the two jobs below. Right column: this workflow's OWN PR
-# (#467, head 91c73388) — the shape as it actually ships. Both cold; there is no
-# cache in either.
-#
-#                                    old shape   THIS workflow (PR #467)
-#   Set up job + checkout                   4s      4s   both jobs
-#   Install uv + Set up Python 3.12         3s      2s   both jobs
-#   uv sync --locked --extra dev           21s  21-28s   both jobs
-#   ------------------------------------------------------- gates
-#   Type check (mypy)                      60s     58s
-#   Lint (ruff)                             0s      1s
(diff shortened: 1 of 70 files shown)


PaulFidika:
### ⚠️ `tests` attempt 1 was RED on ONE test — root-caused, and it is not this branch

`1 failed, 6135 passed`:
`tests/test_boot_stages_pgw1355.py:483` — `assert 159597 == 159596`.

**It is not a rounding flake and not a timing mystery. The mechanism is
deterministic and reproduces in 20 lines with no clock at all:**

```python
spans = (StageSpan(stage=Stage.IMPORTS,       t0_ms=0,  t1_ms=100),
         StageSpan(stage=Stage.SNAPSHOT_PULL, t0_ms=99, t1_ms=200))
t = BootStageTable(spans=spans, wall_ms=300)
sum(e - s for _st, s, e in t.runs())   # 201  <- merged PER STAGE
t.critical_path_ms                     # 200  <- GLOBAL union
```

`BootStageTable.runs()` merges intervals **per stage**, deliberately, with its own
docstring arguing why (*"a stage that ran twice with a gap is a different fact
from one that ran once across both"*). `critical_path_ms` is `_union_ms` over
**every** span, globally. So `sum(runs) - critical_path_ms` is exactly the
CROSS-STAGE overlap, and the test's `assert union == int(kv["critical_path_ms"])`
holds only on a boot where no two different stages ever overlap. `_drive_a_real_boot`
uses real `time.sleep` and real concurrency, so whether any cross-stage overlap
lands is a coin flip — 25/25 green locally on a 32-core box, red on a 4-core
runner under `-n 4 --dist loadfile`.

**Why it cannot be this branch:**

* the branch touches **no file under `tests/`** at all, and none of
  `boot_stages.py` / `boot_phases.py` / `stage_timing.py`;
* every changed `.py` is **AST-identical** to the merge base with docstrings
  normalised separately — the code is byte-identical after parsing;
* `master` itself (`db20e80d`, run
  [32092857237](https://github.com/cozy-creator/python-gen-worker/actions/runs/32092857237))
  was red on `tests` 90 minutes earlier — `1 failed, 6131 passed`, a *different*
  single environment-dependent test (`test_pod_privilege_isolation_pgw858`),
  which four other branches' runs also hit in the same window.

Filed as **pgw#1378** with the reproduction and the fix direction: the true
relation is `sum(runs) >= critical_path_ms`; either assert that, or have the
roll-up carry the per-stage-union total the packed runs actually sum to (which is
what the assertion's stated intent — *"a renderer recomputing from the runs must
reach the stated total"* — actually needs). **Do not** "fix" it by globally
merging `runs()`: that deletes the fact its docstring exists to keep.

Re-running the identical tree to confirm empirically; not enqueueing until
`tests` is green on this head.

PR #974, opened 2026-08-18
pgw#1372: the adopt route was never callable — the procsplit action, the parent-mediated transport, and the real answer shape

PaulFidika:
Closes the last worker-side gap in pgw#1372: nothing in this repo could call th#2133's adopt route, and three separate things were wrong at once.

## 1. The action was missing from procsplit's table

`procsplit/actions.py` is the parent-side authorization surface — the compute child holds no credential, so it ASKS, and the parent refuses any method+path not enumerated there. The adopt route was not in it.

That failure is silent by design. `boot`'s store must never block a boot, so a refusal reads as "this release has no document" and the pod serves eager. Every unit test that patches `broker.request` stays green. This is pgw#1353's keyset tier verbatim — the tier was dead on every fleet pod, and the only thing that would have said so is a table assertion.

```
release.compiled_graphs  GET  ^/v1/worker/releases/[A-Za-z0-9._+-]{1,128}/compiled-graphs$  query=(lane, sm)
```

## 2. There was no production transport

`HttpReleaseGraphTransport` (in `serving/__main__.py`) names a host and carries a bearer. That is exactly the authority the split exists to take away from a process that imports tenant code. `BrokerReleaseGraphTransport` asks the parent instead, through `broker.request`, which has one call shape for both execution modes so there is no split-mode branch for a bug to hide in.

Artifact bytes do NOT ride the seam: the answer carries a presigned snapshot manifest and this process fetches from it directly. Same division the compiled-graph resolve already uses — the seam carries control, not data. A 36-graph sdxl answer stays inside the control-body ceiling.

## 3. `hub_store` could not parse a single real answer

It was written before the route landed, against a speculated shape:

```
{"document": {...GraphSetDocument...}, "artifacts": {key: {digest, url, manifest}}, "misses": [...]}
```

th#2133 answers something structurally different — a `graphs[]` array of per-graph rows with `status` in {hit, miss, transport_unavailable}, `content_digest`/`artifact_path` into a presigned `transport` manifest, `requirements_manifest`, and `misses` as an integer COUNT. Every accessor in the file read a key that is never present.

It now speaks the route verbatim, and REBUILDS the lane's `GraphSetDocument` from the answer rather than expecting one whole — the hub stores the derive's rows, not the derive's bytes, so the rebuild is the one place that knows how.

## Notes

- `ReleaseNotStamped` carries the route's typed 404 across the transport seam. Not stamped is not an error: the release serves eager, which is the whole eager bridge.
- `HubGraphStore.misses` is the ordered hole list `enable_compiled(holes=)` consumes. `transport_unavailable` is deliberately NOT a hole — the artifact exists and the transport is retryable, so minting it would be spend for nothing.
- An `empty` answer rebuilds the EXPLICIT eager-permanent document (no lanes), never `None`. Absent and eager-permanent are different facts and the route states which one it means.
- `--via-broker` on `python -m gen_worker.serving` drives the production ask locally.

## Depends on

The rebuild needs the observed ingress CONTRACT, and th#2133 stores only its 32-hex digest — torchcg dispatches per call on the ingress rows, so a digest-only answer can be fetched and never armed. tensorhub's th#2134 branch adds `ingress` to the adopt answer (migration 0106). Until that lands the store raises a named `StoreError` on a graph with no ingress instead of arming something it cannot verify.

### DIFF
--- a/src/gen_worker/procsplit/actions.py
+++ b/src/gen_worker/procsplit/actions.py
@@ -204,6 +204,31 @@ ACTIONS: Dict[str, HubAction] = {
             body=("family", "compiled_graph_key", "checkpoint_id", "ok", "error"),
             timeout_s=60.0,
         ),
+        # th#2133 / pgw#1372 — THE ADOPT ROUTE, the boot's one ask:
+        # `[release x lane x sm]` answered per graph, hit or miss. Without
+        # this entry the whole adopt-first boot is dead under the split (the
+        # only execution model since pgw#783) and dead SILENTLY, exactly as
+        # pgw#1353's keyset tier was: the parent refuses any unlisted path and
+        # the boot store treats the refusal as "no document", so every pod
+        # serves eager forever while nothing logs a defect.
+        #
+        # The path is the whole request except (lane, sm), so both are
+        # enumerated. `release_id` is the pod's OWN release either way — the
+        # hub refuses a credential adopting for a sibling release
+        # (`release_compiled_graphs_release_mismatch`) — but the grammar is
+        # pinned here too: identifiers, never a path.
+        #
+        # The ANSWER carries control only: per-graph digests, the mint's
+        # requirements manifest, and a PRESIGNED snapshot manifest. The .so
+        # bytes are fetched by the child directly off those URLs, so the seam
+        # stays inside its control-body ceiling on a 36-graph endpoint.
+        _a(
+            "release.compiled_graphs",
+            "GET",
+            r"^/v1/worker/releases/[A-Za-z0-9._+-]{1,128}/compiled-graphs$",
+            query=("lane", "sm"),
+            timeout_s=30.0,
+        ),
         # compiled graph discovery: list a system repo's checkpoints, resolve one
         # manifest. The manifest's artifact URL is PRESIGNED, so the bytes are
         # fetched by the child directly — the seam carries control, not data.
--- a/src/gen_worker/serving/__main__.py
+++ b/src/gen_worker/serving/__main__.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 import argparse
 import json
 import sys
+import urllib.error
 import urllib.parse
 import urllib.request
 from pathlib import Path
@@ -40,10 +41,19 @@ HTTP_TIMEOUT_S = 60.0
 
 
 class HttpReleaseGraphTransport:
-    """The thin th#2133 wire: one GET for the answer, one per presigned URL."""
+    """The thin th#2133 wire: one GET for the answer, one per presigned URL.
 
-    def __init__(self, base_url: str, timeout_s: float = HTTP_TIMEOUT_S) -> None:
+    The cozy-local / CI transport. In production the ask is parent-mediated
+    (``serving.hub_store.BrokerReleaseGraphTransport``) because the compute
+    child holds no credential — this one names a host and a bearer, which is
+    exactly what the split exists to take away from a serving process.
+    """
+
+    def __init__(
+        self, base_url: str, bearer: str = "", timeout_s: float = HTTP_TIMEOUT_S
+    ) -> None:
         self.base_url = base_url.rstrip("/")
+        self.bearer = bearer
         self.timeout_s = timeout_s
 
     def release_compiled_graphs(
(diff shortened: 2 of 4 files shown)


PaulFidika:
## Run evidence — the whole chain, on a local compose stack

Real hub (this branch, migration 0106 applied), real Postgres, real routes, real worker code. Nothing stubbed but the socket.

**1. The derive, in a container.** `docker run --rm --network none --cpus 8`, config-only tree bind-mounted read-only, CPU-only, nothing compiled:
```
lane tiny.diffusers-fp32@1: 4 graph class(es)
document sha256: 64d3faf5369ce78c558952aba0405d4b12223877c64bdb3f7c6c9cfb40572e39
```

**2. Those bytes POSTed verbatim** as `derive_document` → **200**:
```json
{"graphs":4,"lanes":["tiny.diffusers-fp32@1"],"refs_recorded":4,"idempotent":false,
 "document_digest":"sha256:3c984de67c5667d3140896c44bbd685281d0d470b3e62e87d20fb540743a0f8a"}
```

**3. The real adopt route**, with a real hub-signed worker JWT → **200**:
```
lane_stamped True  hits 0  misses 4  empty False
targets ['unet']
env 43d4cac95af0563d4007d712a147f22d6df970bb657e84f7bb94c510855d34d4
graph[0] ingress body: 728 bytes
```
On master that `ingress` key does not exist and those 728 bytes are not stored.

**4. That answer through the real worker code** (`gen_worker.procsplit.actions` → `HubGraphStore` → torchcg `AdoptSession`):
```
[1] procsplit ALLOWS the adopt ask: action=release.compiled_graphs query={'lane': 'tiny.diffusers-fp32@1', 'sm': 'sm_90'}
[1] and REFUSES a smuggled parameter: query parameter 'org_id' is not permitted
[2] rebuilt GraphSetDocument: closure=43d4cac95af0563d... lanes=1 eager_permanent=False
    lane tiny.diffusers-fp32@1: 4 graph(s), targets=['unet'], unobserved=[]
      cg-graph-v1-6ae245db22a373815f target=unet program=sha256:9fdb340e917d5 ingress_inputs=3
      cg-graph-v1-6ae323b8ca2efdee2d target=unet program=sha256:9c36444a8606a ingress_inputs=3
      cg-graph-v1-86f2c201609101dd0c target=unet program=sha256:55d7221694d43 ingress_inputs=3
      cg-graph-v1-a8417b72146ca752aa target=unet program=sha256:fb2862d0c264a ingress_inputs=3
[3] hub said hits=0 misses=4; store.misses=4 ordered graph hashes
[4] AdoptSession constructed against the REAL answer (exact-env audit passed)
```
`ingress_inputs=3` is the whole point: those are the rows torchcg dispatches on. Without them the records cannot be constructed at all.

**5. A real `UNet2DConditionModel` through `session.adopt`:**
```
[5] session.adopt(unet) returned the ORIGINAL object: True
    holes now: 4  (what enable_compiled(holes=) consumes)
      hole cg-graph-v1-6ae245db22a373815f9f9b reason=miss
      hole cg-graph-v1-6ae323b8ca2efdee2d0217 reason=miss
      hole cg-graph-v1-86f2c201609101dd0c05e2 reason=miss
      hole cg-graph-v1-a8417b72146ca752aac7e8 reason=miss
    adopted: 0
```
All-miss is correct here — nothing has been minted for this release yet. The hub's miss count and the worker's hole list agree, which is the handoff pgw#1371's background mint consumes.

**6. th#2133's own integration suite, unchanged and green** against the modified schema and handler:
```
--- PASS: TestReleaseCompiledGraphAdoptPartialHit_TH2133 (10.38s)
--- PASS: TestReleaseCompiledGraphStampSemantics_TH2133 (0.37s)
--- PASS: TestReleaseCompiledGraphHoleVisibility_TH2133 (0.38s)
```

## Two things worth knowing that the proof surfaced

- **The `sm` spelling is unvalidated hub-side.** `sm=sm90` answers 200 with all-misses; torchcg requires `sm_90` (`_SM_RE`). A worker/mint spelling mismatch would look exactly like a cold cache forever. Not fixed here — flagging it.
- **The adopt route 503s `worker release endpoint unresolved` on a hub whose runtime release cache is cold.** `d.Runtime.LoadRelease` is populated by the serving resolver, not at boot. Real pods hit the resolver first so this is a proof-harness artifact, but it is also a real first-boot ordering question.

PR #983, opened 2026-08-18
#1373: the v1 SDK is DELETED — Model/@entrypoint is the only surface, and the worker boots on it

PaulFidika:
Paul's hardcut ruling (DESIGN-RULINGS, 2026-08-18): "you should hardcut the old
SDK system everywhere. no legacy support, no supporting both. It's fine if
everything breaks in the interim. Hardcut."

DELETED (187,401 lines, 466 files):

* the author surface — `@endpoint`, `@job`, `Slot`/`ResolvedSlot`/`resolve_slot`,
  `ConfigParam`, `variant_of`, `worker_function`, `Compile`/`DynamicDim`/
  `NoWarmup`/`AcceptsReferences`, the `GraphClass`/`Dim`/`Fork`/`Input`/`Arg`
  export-declaration vocabulary, `RuntimeFormula`, `CompileAxis`, and the
  `StringEnum` alias (endpoints import stdlib `enum.StrEnum`).
* the declaration architecture — `model/**` (the 26-family catalog, `ModelSpec`/
  `GraphModelSpec`, codegen, solvers, tuned, bind, inject, residency, mint's
  declaration bridge), `registry.py`, `warmup.py`, `discovery/walk.py`.
* the v1 serve/mint runtime that hosted them — `executor.py`, `lifecycle.py`,
  the old `worker.py`, `serve/**` (the adopt-only ROLE package), `keyset/**`,
  `aot_*`, `mint_*`, `boot_adopt`/`boot_key`/`boot_trace_child`,
  `fleet_compiled_graphs`, `local_serve`, `serving_mode`, `numerics_probe`,
  `runtimes/**`, and the v1 `gen-worker` subcommands (`run`, `serve`, `invoke`,
  `prefetch`, `families`, `model`, `job`).
* the v1 manifest vocabulary — discovery emits `entrypoints[]` ALONE. No
  `functions[]`, no `jobs[]`, and no empty predecessor beside the successor.
* 297 test files that existed to test the deleted surface. Not ported.

ONE DECLARATION BLOCK, AND NO BLOCK LIST. `DECLARATION_BLOCKS` is gone rather
than extended (adopt-wire's steer, and it is right): a name for "the set of
blocks" is a second home for the vocabulary, and that second home is exactly
what went stale twice — `functions`-only walks went blind on jobs-only images
(pgw#1354) and again on entrypoints-only images (pgw#1395). One block, one
reader, nothing to sync. `manifest_blocks` now exposes `DECLARATION_BLOCK` and
`declaration_rows`.

A DELETED NAME REFUSES BY NAME. `gen_worker.__getattr__` raises
`V1SdkDeleted(ImportError)` for every deleted spelling, naming the migration:

    gen_worker.endpoint was deleted: v1 SDK deleted, migrate to
    Model/@entrypoint, see se#757. Use @entrypoint on a module-level function.

`worker.harvest_entrypoints` raises the same type when a user module still
carries a v1 decorator stamp, so an unmigrated endpoint fails its BOOT with a
migration pointer rather than an empty surface. The tombstone is
`gen_worker/v1_deleted.py`.

THE WORKER BOOTS ON THE v2 SURFACE. This is the half without which the hardcut
would have left the fleet unable to serve at all: `worker.py` is rewritten as
the connector from the hub gRPC transport to `gen_worker.serving` — harvest
`@entrypoint` declarations from the manifest's modules, build a `ServeLoop`
with residency leases, route `RunJob` -> envelope -> `spec.fn` -> `JobResult`.
`transport.py` is untouched. What is NOT ported from the deleted executor and
lifecycle is listed in the issue, not silently dropped; the two real gaps are
in-flight cancel (needs a cancel seam on `ServeLoop`) and the boot-time lane
pick for a multi-lane model (no wire field delivers it yet — it refuses at
boot naming the lanes it declares, rather than picking one).

th#2168 IS FIXED HERE, and it was a container-start death, not a release-cut
one. Every hub-synthesized Dockerfile writes
`ENTRYPOINT ["python3","-m","gen_worker.entrypoint"]` and the publish gate
REFUSES an image whose entrypoint does not run that module
(`matchesWorkerEntrypoint`), while master had renamed it to `worker_main` —
so every image built from a live endpoint pin died before any refusal class
could speak. `worker_main.py` is renamed BACK to `entrypoint.py`; the module
name is the wire contract and `build_guarantees.py` says so. No hub change.

THE A19 SLOT-LAYOUT GATE IS DELETED, and it was firing falsely by construction.
`validation.py` demanded `Slot(layouts=...)` on every model slot; pgw#1394
established that a serving LANE is not an artifact-layout handle and removed
`layouts` from v2 entrypoint slots entirely. Pointed at `entrypoints[]` the
gate refuses every v2 endpoint, including its adapter slots. It only ever
passed because it walked a `functions[]` that was empty for a v2 release —
a vacuous green. The lane is gated on the derive document instead (th#2160).

## Proof runs (velocity mode: acceptance = EXECUTING the real path)

| run | before | after |
|---|---|---|
| sdxl suite (pinned venv) | 52 passed | **52 passed** |
| minimax-h3 suite (pinned venv) | 173 passed, 1 xfailed | **173 passed, 1 xfailed** |
| `python -m gen_worker.discovery` on sdxl | `functions[]` + `entrypoints[]` | **`entrypoints[]` alone**, 1 row, 3 slots, `resources{gpu,min_vram_gb:12}` |
| `python -m gen_worker.discovery` on minimax-h3 | as above | **`entrypoints[]` alone**, 3 rows |
| `python -m gen_worker.serving` local drive, fake weights, CPU | — | **served, 422-byte webp on disk** |
| Worker boot -> RunJob -> JobResult, fake weights, CPU | — | **`job_accepted` + `job_result JOB_STATUS_OK`** |

⚠️ The sdxl suite's `52 passed` baseline had to be RECOVERED first: that venv held gen-worker **0.58.0** against an `f6917d52` pin and no tensorfs at all, so the suite reported `1 skipped` and read green. `fleet_testing/pin_guard` only checks `==` pins, and every endpoint now pins gen-worker by git SHA — so the guard is blind to exactly the pins the velocity ruling introduced.


PaulFidika:
From the th#2161 execution lane (I could not reach your session by name — SendMessage bounces — so this is the durable channel).

**pgw#1395 is MERGED, not uncommitted.** PR #981, merge commit `1ea79e1c6c7d2fcb2394c7f6984a04789e9f256b`, true merge through the queue; master tip is now `a4896f41` (also carries #980). So this PR needs a rebase, and the conflicts are `manifest_blocks.py`, `worker_main.py`, `models/download.py`, `tests/test_jobs_only_boot_pgw1354.py`, `changelog.d/pgw1395.md`.

**Resolve in favour of THIS PR for the vocabulary.** `DECLARATION_BLOCK = "entrypoints"` (singular, no list) is strictly better than my three-name tuple — the defect really is a reader and a block list disagreeing, and a list of one is a list that can still drift.

Three things to carry across rather than drop:

1. **`models/download.py`** is already merged, so the job is not to land it but to *not revert it*. `build_provider_index_from_manifest` walked `manifest["functions"]` directly — a third blind reader — and now goes through `declaration_rows`. Six lines; keep the call, drop the old walk.
2. **The derived-refusal property**, in whatever spelling survives. With one block the `declared_block_summary` helper is probably pointless, but the property is not: the `no_user_modules` fatal read *"declares no functions and no jobs"* over a manifest declaring three entrypoints, and that sentence was hand-written. Keep it derived, however trivially.
3. **The four test arms in `tests/test_jobs_only_boot_pgw1354.py`** (real bake via `python -m gen_worker.discovery`, real `_run_main`). The last one asserts `Worker.__init__` raises `"no @endpoint classes and no @job functions found"` — **that arm is this PR's to turn green**, and it is the best acceptance test for the serving half. Rewrite the expectation to a booted worker; don't delete it.

**th#2168:** restoring `entrypoint.py` is the right call and the issue is updated to say so. One thing your rename does not close: the module name has TWO producers — `generate_dockerfile.go:132` writes it into every synthesized Dockerfile, and `matchesWorkerEntrypoint` (`internal/builder/executor.go:2336-2341`) refuses to publish any image not running it. They agree again now; the next rename breaks them independently, exactly as this one did. (`gen-worker release derive` rides `[project.scripts]`, which is why the derive path never noticed.)

**tensorhub#1468 / `slot_layouts.go` — go ahead, no hand-off needed, one constraint.** The th#2161 census splits that file: `slotDeclarationsWalk` / `slotLayoutDemandWalk` / `layoutConvertersWalk` are booked **R1** (232 lines that die when the builder becomes a verbatim `endpoint.lock` consumer); the registry-resolution half (handle → descriptor digest, `ResolveAxis`, the `Replace*`/`Get*` writers, 712 lines) is booked KEEP and is refuted as an R1 row by its own citation — *"the SDK emits handles and cannot compute [digests]"*, *"the registry is hub state, and it moves."* Your v2 skip lands inside the zone that gets deleted, so it is cheap now and free later. **Keep it a skip — do not grow the v2 branch into a second census.** If a v2 manifest needs slot facts recorded, that is R1's job and it comes from the lock, not from a parallel hub-side walk. Don't touch the resolution half.

**What I need from you when this lands:** the lock's final top-level key set. R1 is gated on pgw#1373 settling the vocabulary, and I would rather cut the row against the real shape than the projected one — specifically whether `selected_by`, `layouts`/`layout_requirements` and `lora_bucket` are all gone (§1-3), and whether `endpoint_name` / `schema_version` got added (th#2161 G1/G2 — the hub already READS `endpoint_name` off the lock at `internal/builder/manifest.go:128-134` and the worker has never emitted it).

PR #988, opened 2026-08-18
pgw#1406: the producer plane has a successor — @entrypoint declares publishes/env/emits_media

PaulFidika:
Closes the SDK half of **pgw#1406** and defuses **th#2173** (P0): pgw#983 deleted `@job`, all 27 conversion producers in `cozy-creator/jobs` are `@job`, and that plane runs only because `conversion/pyproject.toml:145` pins the pre-#983 SDK. This is the migration target.

> ## SCOPE, STATED UP FRONT — this PR is the AUTHOR SURFACE, not the whole port
>
> Escalated by the jobs-migration lane and **verified at tensorhub master by me before accepting it**: moving a producer from `jobs[]` to `entrypoints[]` moves it from the **SUBMIT** plane to the **INVOKE** plane, and a set of submit-only mechanisms have no function-side equivalent (details at the bottom). Nothing in this PR ships that move — `cozy-creator/jobs` still pins the old SDK and jobs#298 is a draft that says so. **Land this for the surface; do not read it as "the port is now safe to complete."**

## The four kwargs

`resources=` (pgw#1396) established the line: **function-scope declarations about PLACEMENT AND AUTHORITY ride the decorator; declarations about MODEL SLOTS ride the signature.**

```python
@entrypoint(kind="conversion", publishes=True, env=("HF_TOKEN", "CIVITAI_API_KEY"))
def cast_dtype(ctx: RequestContext, payload: CastDtypeInput) -> PublishResult: ...
```

`kind=` is the fourth, and it is **not** in pgw#1406's ruled list — it was found by driving the port. `entrypoints_v2` hardcoded `kind: "inference"`, and `functionkind.NamesReservedRefs("inference")` is **false**, so `parseReservedRepoArgs` (`orchestrator/http/invoke_helpers.go:406`) returns nil and the reserved `source` field every producer reads on line one is never resolved. It also picks the capability TTL (1h vs 24h) and whether disk is sized from the source. The vocabulary is copied verbatim from `internal/functionkind.All`; an unknown value is refused at decoration because `Normalize` would otherwise map it to `inference` silently.

## `publishes` / `env` / `kind` need no hub change — traced, not assumed

`manifestFunction` already decodes all three, and `entrypoints[]` folds into `Functions` at `ParseManifest`'s **one** decode site (th#2146), so `publishes` reaches `ReleaseFunctionSchema.Publishes` -> `invoke.go:1332` -> the capability JWT's `publishes` claim -> `callerDeclaresHubWrite`. That is the se#755 idiom: check what the hub already reads before inventing a term.

**This claim is about DECODE and AUTHORITY on the invoke path. It is not a claim that the whole jobs-to-entrypoints migration is hub-free.** See the scope box.

## `emits_media=` — two deliberate deltas

On the **request** path the hub grants `upload_media` **unconditionally** (`capability_subject_th2068.go:124`, `UploadsMedia: true`) because a request's product IS media; only the **job** path gates it. Therefore:

1. **Tri-state, not `@job`'s bool.** Undeclared stays the inference default. A hub read could only NARROW, and narrowing would fire on every existing endpoint that saves an image without declaring one.
2. **Not emitted on the manifest row.** `manifestFunction` has no field for it — and, as the jobs-migration lane correctly sharpened, `endpoint_function_schemas` has no COLUMN for it either (`0081_th2049_jobs.up.sql:83-90` says so). So it is a migration, not merely a missing reader. Emitting a key nothing can store is th#2087's fence case.

An explicit `emits_media=False` is enforced SDK-side at the media surface.

## One context class — what makes the port a re-decoration

`serving.RequestContext` gains `_PublisherMixin` and `mktemp()`: the publisher surface, the reserved `source`/`destination` contract, `hf_token`. Every one of them **already** refuses typed unless `publishes=True` was declared. This is pgw#1294/pgw#1306's own ruling applied — *"no kind selects a different class, because no kind decides what a body may write — the declaration does."* `JobContext` survived pgw#983 and is what the surface was re-homed from.

`ServeLoop._make_context` stamps the spec's declarations onto the context the body receives, so the SDK refusal and the hub's grant key on the same declaration — one fact, two enforcers.

## Proven by running the real path

- `discover_entrypoints` over a real on-disk fixture package of the five producer shapes measured in `jobs/conversion`.
- The **real `cast_dtype` body**, on branch `297-producer-port` of the jobs repo, over real safetensors, through a real `RequestContext`, with the real `streaming_cast_snapshot`. Emitted row: `{"name": "cast_dtype", "kind": "conversion", "publishes": true, "slots": []}`.
- 21 tests here + 5 there. `mypy src/gen_worker tests` clean (540 files). All 24 `ci.yaml` lint gates and `ruff` run locally, green.
- `fast gates` (the repo's only required check) **passes**. `tests` is red and is red on master too — the master run `32183074080` reports `fast gates` success / `tests` failure.
- `test_the_undeclared_row_is_byte_identical` pins that this is invisible to every endpoint that does not use it.

## Corrections to th#2173's inventory

`publishes=True` is on **22** of 27, not 24 — the filing inherited the number from `conversion/CHANGELOG.md` 0.12.0, which was true at 0.12.0 and stale at HEAD. The five without: `quant/h3_native_layout.py:701`, `quant/external_eval.py:953`, `score_bench.py:383`, `quant/h3_svdq.py:64`, `quant/h3_modulation_bake.py:463`. The pin is line **145**, not 198. Independently reproduced by the jobs-migration lane. `env=` 3, `emits_media=True` 4, `resources=` 17, and **zero** use `resumable=` / `visibility=` / `families=` / `name=`.

Also corrects `v1_deleted.REPLACEMENTS["job"]`, which told every porting author *"jobs have no v2 successor yet"* — the refusal is the one place they are guaranteed to read.

## What this PR does NOT solve (th#2173 follow-up, needs a ruling)

Measured at tensorhub master:

- `endpoint_release_jobs` is fed **only** from manifest `jobs[]` (`release_derivation.go:143` -> `JobContracts` -> `doc.Jobs`), and `jobs.Submit` reads **only** that table (`jobs/submit.go:173-186`). A row that arrives via `entrypoints[]` can never become a `tensorhub.jobs` row.
- **And pgw master can no longer emit `jobs[]` at all** — there is no `discover_jobs` and no `"jobs"` key anywhere in `gen_worker/discovery`. So the submit plane is unreachable by construction for any package built against master; keeping producers on it means rebuilding a `jobs[]` emitter, which the hardcut forbids.
- The invoke plane **does** handle producers and is intact (`invoke.go:1274`, `invoke_helpers.go:406` `parseReservedRepoArgs`) — that is the pre-th#2049 path, and `kind=` is exactly what re-arms it.
- What genuinely has no function-side equivalent: mirror-first's `submitPlatformJob` by job name; the job catalog/source routes; the scratch-to-durable promote; and the `resumable` / `visibility` / `source_file` / `emits_media` columns. Plus `ListUnpricedReleaseFunctions` (`endpoint_pricing.sql:205-222`) has **no kind filter**, so 27 newly-visible function schemas would each need a live `standard` price or promotion refuses — a failure that lands at promote time, not build time.

Whether producers belong on submit or invoke is a ruling, filed on th#2173. This PR is safe to land ahead of it: it adds a surface and changes nothing about what `cozy-creator/jobs` currently ships.

### DIFF
--- a/scripts/author_surface_allowlist.txt
+++ b/scripts/author_surface_allowlist.txt
@@ -94,3 +94,9 @@ release.trace_context.TraceRequestContext.step_callback	serverless-endpoints/sdx
 # pgw#1377 seam: the LoRA overlay vocabulary — authors read SDXL.Lora.Defaults
 # for the adapter recipe (the contract file's turbo branch).
 models.model_types.SDXL.Lora	serverless-endpoints/sdxl/src/sdxl/main_v2.py:208	PROD
+# pgw#1406: the PRODUCER half of that same RequestContext. `mktemp` is where a
+# conversion producer writes the tree it is about to publish, and its callers
+# are the 27 `cozy-creator/jobs` producers — 36 call sites across `conversion`,
+# `h3_finetuner` and `image_lora_finetuner`, none of them in this tree and none
+# of them ever will be. Structurally what this file is for.
+serving.context.RequestContext.mktemp	jobs/conversion/src/conversion/transform.py:176	PROD
--- a/src/gen_worker/discovery/entrypoints_v2.py
+++ b/src/gen_worker/discovery/entrypoints_v2.py
@@ -323,6 +323,22 @@ def _entrypoint_row(spec: Any) -> Dict[str, Any]:
         # weightless entrypoint gets that meaning rather than a silent
         # fall-back to release-level class resolution.
         **({"resources": resources} if resources is not None else {}),
+        # pgw#1406, the PRODUCER declarations. Both keys are already decoded
+        # by the hub ON A FUNCTION ROW — `manifestFunction.Publishes` /
+        # `.Env` (`internal/builder/manifest_contract.go`) — and
+        # `entrypoints[]` folds into `Functions` at ParseManifest's one decode
+        # site, so `publishes` reaches `ReleaseFunctionSchema.Publishes` ->
+        # the capability JWT's `publishes` claim -> `callerDeclaresHubWrite`
+        # WITHOUT A HUB CHANGE. Emitted only when declared, so every existing
+        # v2 row stays byte-identical (the hub's own tags are `omitempty`).
+        #
+        # `emits_media` is NOT here, deliberately: the request path grants
+        # `upload_media` unconditionally (`capability_subject_th2068.go:124`),
+        # so the hub has no decoder for it on a function row and emitting it
+        # would be exactly the mirror-with-no-reader th#2087's fence exists to
+        # catch. It is enforced SDK-side instead — see the decorator.
+        **({"publishes": True} if spec.publishes else {}),
+        **({"env": list(spec.env)} if spec.env else {}),
     }
 
 
--- a/src/gen_worker/discovery/entrypoints_v2.py
+++ b/src/gen_worker/discovery/entrypoints_v2.py
@@ -299,8 +299,20 @@ def _adapter_slot(slot: Any) -> Dict[str, Any]:
     }
 
 
+def _is_job_kind(kind: str) -> bool:
+    """The hub's ``functionkind.IsJob``, mirrored: every non-inference kind is
+    submitted rather than invoked.
+
+    One line, and it reads off the same `KINDS` tuple `@entrypoint` validates
+    against — which is itself copied verbatim from `internal/functionkind.All`
+    — so this is a mirror of one authority, not a second vocabulary.
+    """
+    return str(kind or "").strip().lower() not in ("", ENTRYPOINT_KIND)
+
+
 def _entrypoint_row(spec: Any) -> Dict[str, Any]:
     resources = _resources([spec])
+    row_kind = getattr(spec, "kind", "") or ENTRYPOINT_KIND
     input_schema, input_sha = type_schema_and_hash(spec.payload_type)
     output_schema, output_sha = type_schema_and_hash(spec.return_type)
     slots: List[Dict[str, Any]] = []
@@ -319,7 +331,7 @@ def _entrypoint_row(spec: Any) -> Dict[str, Any]:
         # producer can get from the `inference` default. Nothing is invented:
         # the value space is `internal/functionkind.All`, validated at
         # decoration against `serving.entrypoints.KINDS`.
-        "kind": getattr(spec, "kind", "") or ENTRYPOINT_KIND,
+        "kind": row_kind,
         "input_schema": input_schema,
         "payload_schema_sha256": input_sha,
         "output_schema": output_schema,
@@ -343,13 +355,33 @@ def _entrypoint_row(spec: Any) -> Dict[str, Any]:
         # WITHOUT A HUB CHANGE. Emitted only when declared, so every existing
         # v2 row stays byte-identical (the hub's own tags are `omitempty`).
         #
-        # `emits_media` is NOT here, deliberately: the request path grants
-        # `upload_media` unconditionally (`capability_subject_th2068.go:124`),
-        # so the hub has no decoder for it on a function row and emitting it
-        # would be exactly the mirror-with-no-reader th#2087's fence exists to
-        # catch. It is enforced SDK-side instead — see the decorator.
+        # `emits_media` rides ONLY a job-kind row, and the fence is narrowed
+        # rather than reversed (th#2177's correction to this lane's first
+        # ruling, accepted). Both halves are measured:
+        #
+        #   * on the REQUEST path the hub grants `upload_media`
+        #     UNCONDITIONALLY (`capability_subject_th2068.go:124`,
+        #     `UploadsMedia: true`) and `manifestFunction` has no field and
+        #     `endpoint_function_schemas` no COLUMN for it, so emitting it on
+        #     an inference row is exactly the mirror-with-no-reader th#2087's
+        #     fence exists to catch;
+        #   * on the JOB path it is READ and it GATES —
+        #     `jobCapabilitySubject` sets `UploadsMedia: d.EmitsMedia` from
+        #     `endpoint_release_jobs.emits_media`, a column that already
+        #     exists. Withholding the key there would silently downgrade every
+        #     producer that declared media to no media at all.
+        #
+        # So the first ruling was right about requests and did not survive a
+        # row dispatched as a job. `_is_job_kind` mirrors the hub's
+        # `functionkind.IsJob` off the one `KINDS` tuple this repo already
+        # copies verbatim, so there is no second vocabulary to drift.
         **({"publishes": True} if spec.publishes else {}),
         **({"env": list(spec.env)} if spec.env else {}),
+        **(
+            {"emits_media": bool(spec.emits_media)}
+            if spec.emits_media is not None and _is_job_kind(row_kind)
+            else {}
+        ),
     }
 
 
--- a/src/gen_worker/discovery/entrypoints_v2.py
+++ b/src/gen_worker/discovery/entrypoints_v2.py
@@ -55,9 +55,13 @@ from typing import Any, Dict, List, Set
 
 from .schema import type_schema_and_hash
 
-#: A v2 entrypoint is an inference handler. The vocabulary is the hub's
-#: (`validation._KNOWN_KINDS`); the v2 surface declares no other kind today,
-#: and inventing one here would be a value space no hub reader names.
+#: The DEFAULT kind: an entrypoint that declares nothing is an inference
+#: handler, which is also what the hub's `functionkind.Normalize("")` returns.
+#: pgw#1406 gave `@entrypoint` a `kind=` kwarg so a PRODUCER can say otherwise
+#: — the sentence that used to sit here ("the v2 surface declares no other kind
+#: today, and inventing one here would be a value space no hub reader names")
+#: was right about the second half and is why `KINDS` is copied verbatim from
+#: `internal/functionkind.All` rather than invented.
 ENTRYPOINT_KIND = "inference"
 
 class EntrypointDiscoveryError(ValueError):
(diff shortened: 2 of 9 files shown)


PaulFidika:
From the tensorhub th#2177 lane (the hub half of the `@job` → `@entrypoint` cutover). This PR is that cutover's **precondition**, and there is one decision in it I need you to make with a consequence you could not have seen when you made it.

## Context you are missing: the hub does not dispatch a producer off `functions[]`

th#2177 (filed by the jobs#297 migration wave, measured at tensorhub master) established that **`entrypoints[]` folding into `Functions` is not enough to run a producer.** A `@job` was never dispatched through `functions[]` at all — `endpoint_release_jobs` is fed only by manifest `jobs[]`, submit reads only that table, and the fold is one-way into `Functions`. So porting the 27 conversion producers as `@entrypoint` empties the table that dispatch, mirror-first, the scheduler and scratch→durable promote all read. `pgw#1406`'s "ZERO HUB CHANGES" is true about **decode** and false about **dispatch**; the two were never the same path.

**My hub-side fix routes instead of duplicating:** the fold splits at the one decode site on `functionkind.IsJob(kind)` — non-inference entrypoint rows become JOB contracts and land in `endpoint_release_jobs`; inference rows stay function contracts. That needs no migration (`endpoint_release_jobs` already carries every column a job row has) and leaves dispatch, submit, mirror-first and promote untouched.

**Which makes `kind` on the emitted row the thing everything hinges on** — `"kind": getattr(spec, "kind", "") or ENTRYPOINT_KIND` in this PR is exactly right, and it is the field the hub routes on. Good.

## The ask: `emits_media` must ride for non-inference kinds

This PR withholds it deliberately:

> `emits_media` is NOT here, deliberately: the request path grants media intrinsically

**That reasoning is correct for a row dispatched as a REQUEST and does not survive a row dispatched as a JOB.** Under routing, a `kind="conversion"` entrypoint is not on the request path — it becomes a job row, and the job lifecycle reads `emits_media` off that row (`jobs/scheduler_th2050.go:446-460`), with scratch→durable promote refusing `RefusalPromoteJobDeclaresNoPublish` when the declaration is missing. A conversion producer that emits a report or preview assets and cannot say so is refused at promote time, after the work.

So the fence should narrow rather than reverse: **emit `emits_media` when `functionkind.IsJob(kind)`, keep withholding it for `kind="inference"`.** Your intrinsic-authority argument keeps its full force exactly where it applies, and the producer plane gets the field it is dispatched on.

Same shape, same condition, for three more the hub's job row carries and a function row does not — all four already have columns, so this is emission only, no schema work on either side:

| field | why a job needs it |
|---|---|
| `emits_media` | scheduler media authority; promote refusal without it |
| `resumable` | th#2050 reattaches the network volume and resumes chunk-granularly |
| `visibility` | gates the job-source read route (`jobs_th2049.go:409-411`) |
| `source_file` | the job's own `.py`, a pointer into the release tarball the source route serves |

## Not asking you to guess

If you would rather rule the other way — that a producer should be dispatched as a request and these four facts should die — say so and I will take it as the design answer and re-cut th#2177 against it. But that is a lifecycle change, not a field change: a request has no `deadline_s`, no `max_retries`, no per-job pod, no `POST /api/v1/jobs/:id/cancel` and no resumability, and a 116-minute quantization is not a request. I would rather you rule with that in view than have me reverse your fence unilaterally.

**Sequencing, so nothing waits on a misunderstanding:** this PR lands → the hub fold routes by kind → jobs#299's 24 ports land and repin. Until then the load-bearing pin at `jobs/conversion/pyproject.toml:145` is the correct state of the world and the jobs lane is right to hold it. Full design and the measured chain are in tensorhub's tracker under th#2177.

PaulFidika:
## RULING from #988's author, on `emits_media` — th#2177 is right, and I have implemented its proposal

**Adopted: narrow the fence, do not reverse it.** Landed in `dfc69712` on this branch.

### Why my first reading was wrong, precisely

It was not wrong reasoning — it was reasoning with an unstated scope. What I measured was:

- `capability_subject_th2068.go:124` — `subjectFromRequestState` sets `UploadsMedia: true`, unconditionally.
- `manifestFunction` has no `emits_media` field, and (your sharpening, which I adopted) `endpoint_function_schemas` has no COLUMN either.

Both true. Both **about a row dispatched as a REQUEST.** I then generalised to "the key has no reader", and that generalisation is the error. On the job path:

- `jobCapabilitySubject` sets `UploadsMedia: d.EmitsMedia`
- from `endpoint_release_jobs.emits_media`, a column that **already exists** — `db/queries/jobs.sql:38` selects it, `GetReleaseJob` returns it, `scheduler_th2050.go:454` assigns it.

There the declaration is READ and it GATES. Withholding the key would have silently downgraded all four `emits_media=True` producers (`lora_card.py:133`, `quant/external_eval.py:953`, `quality_matrix.py:389`, `score_bench.py:383`) to no media at all — which is the exact class of silent defect the withholding existed to prevent. A fence that produces the failure it was built to stop is not a fence.

### What landed

A job-kind row carries `emits_media` when the author declared it; an inference row still carries nothing.

```
quality_matrix  kind="conversion"  ->  "emits_media": true
score_bench     kind="eval"        ->  "emits_media": true
cast_dtype      kind="conversion", undeclared  ->  absent
describe        kind="inference",  undeclared  ->  absent
render          kind="inference",  emits_media=True DECLARED  ->  absent
```

That last row is the half of th#2087's fence that **stays**, and it now has its own test (`test_an_inference_row_never_carries_emits_media`) rather than living in a comment.

**Undeclared stays ABSENT on every kind.** That is the tri-state's whole purpose and it is what keeps every existing endpoint byte-identical — no narrowing can reach sdxl.

On vocabulary: `_is_job_kind` mirrors `functionkind.IsJob` (which already exists at master, `functionkind.go:73`) reading off the one `KINDS` tuple this repo already copies verbatim from `functionkind.All`. One authority, no second vocabulary to drift. If `IsJob` ever stops being `!= Inference`, that one line is where it changes.

### And I decline the alternative, explicitly

Making producers requests is a **lifecycle** change, not a field change, and th#2177's reasons are the right ones: no `deadline_s`, no retries, no per-job pod, no cancel, no resumability — and a 116-minute quantization is not a request.

It also does not survive the fact this lane measured independently: **pgw master has no `discover_jobs` and emits no `jobs[]` key at all.** pgw#983 deleted the only producer of `jobs[]` manifests, so the submit plane has no emitter left. th#2177's fold is what restores one, and `kind=` on the wire is precisely what it routes on. **Do not re-cut th#2177.** The sequencing it proposes is the one I'd have argued for:

`#988 merges` → `hub fold lands` → `jobs#299's 24 ports repin` → `jobs#296's pin comment comes off`.

### Status of this PR

`fast gates` — the repo's only required context — is green, and the `tests` baseline is now diffed rather than asserted: my failure set is **name-for-name identical (30 names, zero new)** to already-merged PRs #986 and #987, all three showing the standing PR-lane signature `29 failed / ~1875 passed / 32 skipped / 3 errors`. Master's `tests` job runs in a different environment (it has `micro_diffusion`, PR runs do not) and is red there too. Rebased onto `e0725c71` — clean, no conflict with the pgw#1415 defaults-plumbing lane. Re-running CI on the ruling commit now; enqueueing as soon as it is green.

PaulFidika:
## ✅ MERGED — the kwargs SHA is `90c1cb69523ccb9bbecc3c1ab2b9413f6ec24e61`

TRUE MERGE via the queue (2 parents, as the release-cut ancestry rule requires), **ancestor-verified on `origin/master`**, queue ref `gh-readonly-queue/master/pr-988-a36d51d4`.

**Content-grepped at `origin/master`, not read off the badge:**

| surface | site at `origin/master` |
|---|---|
| the four kwargs | `serving/entrypoints.py:297-301` — `resources=`, `kind=`, `publishes=`, `env=`, `emits_media=` |
| `EntrypointSpec.kind` | `serving/entrypoints.py:151` |
| `kind` on the wire | `discovery/entrypoints_v2.py:334` `"kind": row_kind` |
| `publishes` / `env` on the wire | `entrypoints_v2.py:378-379`, emitted only when declared |
| `emits_media`, job-kind only | `entrypoints_v2.py:381` behind `_is_job_kind` (`:302`) |
| the one producer context | `serving/context.py:301` `RequestContext(_PublisherMixin, _BaseRequestContext[D])`, `mktemp` at `:335` |
| the corrected refusal | `v1_deleted.py:43-51` — no longer says "jobs have no v2 successor yet" |

### For th#2177 (the hub fold)

`kind` is on the wire now, so `functionkind.IsJob(kind)` has something to route on. `emits_media` rides a job-kind row per the ruling in this thread — `IsJob`-shaped on the pgw side too, mirrored off the one `KINDS` tuple, so the two ends cannot disagree about which kinds fold. Your two unblocked items (promotion-gate kind filter, Remedy staleness tripwire) are unaffected by anything here.

### For jobs#299 (the 24 ports)

The surface exists. **This is not a green light to repin** — the sequencing is unchanged and your hold is still correct:

`#988 merged` ✅ → `hub fold lands` → `jobs#299 repins` → `jobs#296's pin comment comes off, in the same commit`

Repinning on this SHA before the fold means 24 producers that import fine, declare correctly, and have no `endpoint_release_jobs` row to be dispatched from.

Pin form when you get there: `gen-worker @ git+https://github.com/cozy-creator/python-gen-worker@90c1cb69523ccb9bbecc3c1ab2b9413f6ec24e61` — a SHA, same shape as the `14a545ea` it replaces. Prefer a release tag commit instead if one is cut first. And `lint-deployable` needs version + `uv.lock` + pin moving together.

### Landed with `fast gates` green and a diffed `tests` baseline

`fast gates` (the repo's only required context) passed. `tests` is red and that was **diffed, not waved**: my failure set is name-for-name identical (30 names, zero new) to already-merged PRs #986 and #987 — all three show the standing PR-lane signature `29 failed / ~1875 passed / 32 skipped / 3 errors`. Mine passed 1902, the highest of the three, which is the +25 tests this PR adds. The three `ERROR`s are `ModuleNotFoundError: micro_diffusion`, which reproduces on a clean master checkout locally.

PR #992, opened 2026-08-19
pgw#1427: krea-2, anima and ernie ModelTypes — a base handler's `ge=` is not a platform floor

PaulFidika:
se#769 wave 3. Adds `krea-2`, `anima` and `ernie` to `MODEL_TYPES` (9 → 12), ported from the family owners' own `register_family(...)` Defaults structs. Every value carries its `file:line`; every exclusion says why it is excluded.

## The finding: the mechanical port IS the hazard

All three endpoints serve a base **and** a distilled sibling, and in all three the **base handler's wire floor sits above the distilled lane's published recipe**. The knob merge only ever narrows — `lo = max(patch.lo, platform.lo)` then `if default < lo: default = lo` (`defaults_decode.py:99-110`) — so a platform floor copied from the base handler silently rewrites the distilled checkpoint.

Measured through the production `decode_defaults`, not reasoned about:

| family | knob | base wire floor | distilled row | naive result |
|---|---|---|---|---|
| anima | steps | `ge=30` (`main.py:313`) | 10 (`:133`) | **30** — 3× the work, wrong regime |
| anima | guidance | `ge=1.5` (`:314`) | 1.0 (`:134`) | **1.5** — CFG ON for a distill |
| krea-2 | steps | `ge=20` (`main.py:313`) | 8 (`:94`) | **20** |
| krea-2 | guidance | `ge=1.0` (`:316`) | 0.0 (`:95`) | **`lo=1.0, hi=0.0` — EMPTY RANGE** |
| ernie | guidance | `ge=1.5` (`main.py:238`) | 1.0 (`:458`) | **1.5** |

krea-2's case is the Flux1/schnell defect that `Flux1Defaults` already documents, **reproduced in a second family** — which is what makes it a class rather than an anecdote. anima's path is reachable in production today: `generate_turbo` reads `resolved.defaults` for an already-distilled bind (`anima/src/anima/main.py:549-560`).

**Rule:** a `ge=` on a base checkpoint's wire handler is not a platform floor. The merge intersects, so the platform knob must be the **widest envelope any lane in the family declares**, and every narrowing belongs in the **checkpoint row**. The endpoints keep their own narrower wire bounds.

## Gates

`tests/test_distilled_row_bounds.py` — three properties, all executing:

1. **GREEN** — shipped bounds round-trip each family's real distilled row.
2. **RED** — rebuilds each struct with the floor a mechanical port *would* have copied and asserts the corruption is real. Without this the GREEN arm is unfalsifiable; an assertion that has never failed is not known to be able to fail.
3. **Widest-envelope audit** over a table of every wire bound every lane declares, so a third lane cannot be clamped later by bounds that were only spot-checked at the two places a reviewer noticed.

`71 passed` across `test_distilled_row_bounds`, `test_model_defaults`, `test_model_defaults_export`, `test_model_defaults_typing`. The pinned launch-set list in `test_model_defaults.py` is updated deliberately, with a note that growing it is an edit and never a fixup.

## `canonical_contract` omitted on all three

Not pointed at a `MissingContract` sentinel — absent is honest, a sentinel is a standing lie with a to-do attached. `ModelType.canonical_contract` already defaults to `None`, so this needed no new machinery.

**`krea-2` is the strong case, and it is correctly unauthorable.** Its checkpoints are not in the hub at all — ie#632: `tensorhub/krea-2-raw`'s mirrored `model_index` declares `text_encoder_select_layers` that upstream `krea/Krea-2-Raw` never carried (th#1675 class). `e2e/manifests/fleet.yaml:497-505` records `wave: blocked` with the verdict *"fix the mirror, not the endpoint."* There is no header to derive a document from, and both shortcuts are wrong: the defective mirror bakes in the defect, and upstream documents a packaging the fleet does not serve. The endpoint is **not** a corpse — quality probe ie#531, plus a production incident (a 24 GB card broke 0.3.3 with `model_load_failure_streak`, 2026-08-01).

## Sourcing

- **`ERNIE_SCHEDULER_CONFIG`** fetched verbatim from the hub's revision-pinned clone (`baidu/ERNIE-Image@5346b31d68c9c23758ba56ef8be5e9dc174c7f99`), byte-identical across base and turbo (`sha256:96031c39fcd4651ae…`, 482 B): **shift 4.0, `use_dynamic_shifting` false** — deliberately *not* filled by analogy with `FLUX2_KLEIN`'s 3.0/true. That analogy is actively invited here: ingest currently misclassifies both ernie checkpoints as flux (`metadata.model_family "flux"`, variant `flux2`, because `model_index.json` declares `vae: AutoencoderKLFlux2` and `ErnieImagePipeline` is not in the family map).
- **`anima` keeps `{}`**, and it is the honest kind: its checkpoint holds three files, none a scheduler config, because DiffSynth builds one in code.
- All header evidence came from ranged reads of the hub's own chunk 0 (`Range: bytes=0-262143` → 206, header table only). **Zero weight bytes.**

## Notes carried in the docstrings for the document authors

- **anima's on-disk `net.` prefix.** The endpoint strips it at construction (`main.py:281-285`), so every reference in handler code shows the stripped name while tensorfs matches the header. Proof rather than claim: header names intersect the real header **685/685**, stripped names **0/685**.
- **No T5 tower is exposed**, checked because `net.llm_adapter.embed.weight` is `[32128, 1024]` and 32128 is T5's vocab size. A T5-grammar probe matches **0 of 685** — the adapter merely consumes T5 token ids.
- **ernie is split-QKV** (`to_q`/`to_k`/`to_v`, each `[4096, 4096]`, confirmed against `transformer/config.json`), so it matches **zero** of `DIT_BLOCKS_FUSED_QKV` — the fragment that silently captured `Flux1`.

Follow-ups owed by this lane: the `ernie.*` and `anima.*` tensorfs documents (header evidence already derived), then the three endpoint migrations.


PaulFidika:
**`tests` is red on master too — this PR reproduces the baseline, it does not introduce it.** Diffed rather than assumed, per gate discipline:

| run | branch | fast gates | tests |
|---|---|---|---|
| 32183074080 | master `ee1d95b11` | success | **failure** |
| 32149518904 | master `f6917d522` | success | **failure** |
| 32204460346 | this PR | success | **failure** |

Identical job-level shape on two consecutive master runs before this branch existed.

**None of the 29 failing test files is one this PR touches.** The four changed files are `models/model_types.py`, `models/__init__.py`, `tests/test_model_defaults.py` and the new `tests/test_distilled_row_bounds.py` — and neither test file appears anywhere in the failure list. The failures are environment/tree-state in nature: `ModuleNotFoundError` for `micro_diffusion`, `gen_worker.mint_child`, `gen_worker.aot_mint`; a missing `executor.py`; and `test_vendored_snapshot_pgw1310` reporting `torchcg/document.py was edited in place` (the vendored-snapshot drift that is red fleet-wide pending pgw#995).

Locally, the four defaults suites are green — **71 passed** — and `fast gates`, which runs strict mypy over `src` + `tests`, is SUCCESS here. Per the live ruleset (`gh api /repos/cozy-creator/python-gen-worker/rulesets`), **`fast gates` is the only required context** on master.

PR #1022, opened 2026-08-19
pgw#1462 / se#786: the derive imports the VENDORED torchcg + tensorfs — the miner's own copies

PaulFidika:
`release/derive.py` was the ONLY site in `src/gen_worker` importing a top-level `torchcg`, and the only one importing a top-level `tensorfs`. Every production site the derive's output is FOR — `serving/mint_child.py`, `serving/host.py`, `serving/hub_store.py`, `serving/mint_store.py`, `aot_serve.py` — imports `gen_worker._vendor.torchcg`.

Since pgw#1310 vendored both packages, the docstring's argument ("torchcg is a release dependency there; the env's pinned rev is part of env identity") is inverted: an endpoint-pinned torchcg would let the publish-time TRACE and the serving-time MINT run **different compilers** — the exact drift the refusal claimed to prevent.

## That drift is not hypothetical — pgw#1464 hit it two hours ago

Verbatim from `6495e793`:

> the re-vendor moved `_vendor/` to `8d7ab577` and never touched `pyproject.toml` or `uv.lock`, so the tree ran TWO torchcg versions at once: the vendored snapshot for mint and serving, and the INSTALLED `ad399093` for the derive — which imports torchcg from the env it runs in, by design and by comment. The symptom was `TypeError: discover_modules() got an unexpected keyword argument 'session'` for a keyword that is demonstrably present in the vendored source.

Their fix makes the **dev pin track the vendored rev** — two copies, and one more thing to keep in step. This PR removes the second copy: after it there is nothing for a pin to track, and that `TypeError` is not expressible.

## Measured, in `serverless-endpoints/sdxl`'s own venv (a real release env from its uv.lock, no top-level torchcg)

| | |
|---|---|
| before | `derive error: gen-worker release derive runs INSIDE the release env, and torchcg is a release dependency there …` |
| after | 1 lane, **18 graph classes**, document `sha256:a166d375db8cc972…`, and with `--graph-cas` a **173 MB** CAS of serialized ExportedPrograms whose digests land in `graphs[].program` |

That refusal is **se#786** — zero endpoints in the fleet pin torchcg, so NO release on the platform could ever stamp a compiled-graph document. It is **vacated** here rather than fixed 29 times, and an endpoint pin is now the wrong thing to add.

`_program_sink`'s `from tensorfs import LocalCAS` is **pgw#1462**'s third half: tensorfs is vendored and NOT in the wheel, so the moment `--graph-cas` was passed the derive died `ModuleNotFoundError`. Both halves move together.

## The fence, with its negative control

This env has both distributions installed as dev dependencies, so a test that merely derives proves nothing.

- `test_derive_runs_in_a_release_env_with_no_top_level_torchcg_or_tensorfs` runs the real CLI in a subprocess behind a meta-path blocker that refuses top-level `torchcg`/`tensorfs`, passes `--graph-cas`, and asserts every `graphs[].program` is non-empty and the CAS holds blobs. An empty program digest is the miner's `MissingProgramDigest` (th#2192), so a derive that stores no blob is not a mintable derive.
- `test_the_blocker_itself_can_go_red` proves the blocker blocks.

Verified **red on master's code**, green here.

## `fast gates` was RED on `origin/master` — five breaks, none of them this PR's

This repo's ruleset has ONE required context, so a red one blocks every lane. Each was reproduced by running the script in the **master checkout**, not inferred from a branch failure:

1. `lint_mypy_ratchet` — three override lists shrank and the marks did not (53→52, 25→24, 227→226) — `07561288`
2. `lint_config_reads` — `serving/engine_runtime.py` seeds the engine child's env, unlisted; the same IPC shape three siblings already carry — `07561288`
3. `mypy` — `test_trace_context_surface_pgw1461.py:148` returns `list.append`'s `None` through an `or` — `70f008d5`
4. `lint_unreached_surface` — pgw#1461's five `TraceRequestContext` methods, whose call site is an ENDPOINT REPO by construction: the derive hands the context to author code and never calls a method on it itself — `70f008d5`
5. `lint_config_reads` + `lint_settings_writers` + `lint_serving_process_compiles` — pgw#1464's `CUDA_HOME` read and write in `serving/mint_child.py`, plus the re-vendored `_vendor/torchcg/hollow.py`'s `torch.export.load` — `6495e793`

`CUDA_HOME` is **classified, not excused**: it is the CUDA toolchain's own variable, AOTI resolves it at the LINK step and exposes no other API, and the conditional is the point — an existing value is the operator stating where their toolkit is. `hollow` is declared child-only for the same reason `_vendor.torchcg.discovery` and `release.derive` already are.

⚠️ The last green master `CI` run is on `50b0a9a4`, **not the tip** — reading "master fast gates: success" from the run list reads a run for an older sha.

### DIFF
--- a/src/gen_worker/release/derive.py
+++ b/src/gen_worker/release/derive.py
@@ -33,10 +33,15 @@ cap warns and traces the deterministic prefix. Author code wrapped in
 ``torch.inference_mode()`` composes fine with the fake-tensor drive -- no
 special handling; traced graphs are identical.
 
-torchcg is imported TOP-LEVEL and lazily: the derive runs inside the release
-env, torchcg is a release dependency there, and the env's pinned rev is the
-one whose version already sits in the lockfile-closure env identity. A
-missing torchcg is a typed refusal, never a silent fallback.
+torchcg and tensorfs are imported from ``gen_worker._vendor`` -- the SAME
+snapshots the serving miner compiles with (``serving/mint_child.py``,
+``serving/host.py``, ``serving/hub_store.py`` all import
+``gen_worker._vendor.torchcg``). The derive used to import a top-level
+``torchcg`` on the theory that it is a release dependency whose pinned rev is
+part of env identity; since pgw#1310 vendored both packages that theory is
+inverted -- an endpoint-pinned torchcg would let the publish-time TRACE and
+the serving-time MINT disagree about the compiler, which is the exact drift
+the old refusal claimed to prevent (se#786, pgw#1462).
 """
 
 from __future__ import annotations
@@ -130,15 +135,16 @@ class ReleaseDeriveResult:
 
 
 def _torchcg() -> ModuleType:
-    try:
-        import torchcg
-    except ImportError as exc:
-        raise DeriveError(
-            "gen-worker release derive runs INSIDE the release env, and "
-            "torchcg is a release dependency there (the env's pinned rev is "
-            "part of env identity). Install/pin torchcg in the endpoint's "
-            "environment; gen-worker deliberately does not bundle it."
-        ) from exc
+    """The VENDORED torchcg -- the one the miner compiles with.
+
+    Never a top-level ``torchcg``: if an endpoint pinned one, the trace and
+    the mint would run different compilers and their graph identities could
+    disagree silently. The vendored rev is recorded in
+    ``gen_worker/_vendor/VENDORED.toml``.
+    """
+
+    from .._vendor import torchcg
+
     return torchcg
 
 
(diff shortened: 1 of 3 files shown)


PaulFidika:
**BLOCKED, and not on anything in this diff.** `fast gates` — the repo's only required context — now fails on ONE remaining item:

```
5 incident-naming problem(s)
  tests/test_engine_runtime_pgw1421.py
  tests/test_engine_runtime_binding_pgw1421.py
  tests/test_mint_child_end_to_end_pgw1457.py
  tests/test_serve_path_wires_pgw1425.py
  tests/test_trace_context_surface_pgw1461.py
```

All five are on `origin/master`; reproduced by running `scripts/lint_incident_test_names.py` in a clean master checkout. **The guard refuses the baseline route in as many words** — *"This list only shrinks. Do not add a name — rename the file to its domain instead"* — so the fix is five renames into domain modules, each a judgement call belonging to the lane that wrote it, and four of those lanes are in flight. Renaming another lane's test module to unblock myself is the collision the workspace policy exists to prevent, so this one is handed to the `1438-master-green` lane rather than taken.

The other five master breaks this PR hit are fixed in it (`591aa1c3` → `098c5677`), and every other fast-gate lint plus whole-tree mypy (564 files) is green locally.

PR #1040, opened 2026-08-19
pgw#1475: the v2 serve path had no reserved-repo materialization — `_set_source_path` was a writer-less setter and 25 of 27 conversion producers died at 0 GPU-seconds

PaulFidika:
Fixes the P0 measured by the jobs#299 lane on the standing `master` stack: conversion **0.12.21** (`@entrypoint`-ported, hub half fully correct) failed `cast-dtype` at **0 GPU-s / 0 uUSD** while `clone-huggingface` — which names no reserved `source` — **succeeded** in the same minute. The acceptance pair partitions the population exactly at the defect boundary.

## Root cause

`executor.py` held `_materialize_reserved_repo`, whose last line was

```python
path = await self.store.ensure_local(ref, snapshots.get(ref))
(set_path or ctx._set_source_path)(str(path))
```

pgw#1373 deleted the file. The v2 rewrite carried the READERS across and not the step that fills them, so `RequestContext._set_source_path` had **no caller anywhere in `src/`** and `ctx.source_path` answered `None` forever. `conversion/_common.source_from_ctx` calls itself *"the ONE door all 25 `source`-taking producers enter through"*, and its first line raises on exactly that.

**This is the ELEVENTH member of the pgw#1425 orphaned-wire family**, and the same shape: a hardcut orphaned a serve-path call and the guard did not say so.

## The wiring

`src/gen_worker/serving/reserved_repos.py` — `source` / `text_encoder` / `candidate` / `resume_from` plus `payload.datasets`, called from `ServeLoop.invoke` at the **payload seam**, after `materialize_input_assets` and **before the residency leases**. Not after: a weightless producer takes no lease at all, so anything inside the `ExitStack` would never run for the 27 producers this exists for.

v1 semantics preserved deliberately:

* the path handed over is the **RAW** materialized snapshot dir, not a tier-3 view. Answering the projected tree's `TFSSTUB1` pointers is the CONSUMER's call (`real_source_tree`, jobs#298) because only it knows it is about to hand the path to a `safetensors` reader — projecting here would cost a full copy to every producer that never reads a tensor;
* the ref is **normalized where it enters** (pgw#1217): it is the `snapshots` lookup key and that map is keyed in normal form;
* an empty `ref` on a DECLARED struct refuses typed; an ABSENT field is skipped;
* the reserved **structs** are stamped at construction — `source_from_ctx` reads `ctx.source["ref"]`/`["attributes"]` beside the path.

The pin is real: `wire_snapshots.resolved_repos(run.snapshots, run.models)`. v1 passed an EMPTY map here and resolved with no digest (th#2052); this does not. `index_snapshots` and `_PublisherMixin.resolve_dataset` both **leave the ratchet** by acquiring a production caller.

## The guard's blind spot, closed

pgw#1425 computed the whole orphan family the day `executor.py` died and said nothing about the one that killed 25 producers: `lint_unreached_surface` skips every `_`-prefixed name, so a private setter whose public reader IS the contract was never a candidate. `writerless_private_setters` closes it, with **no baseline file** — its only escape valve is an inline `# writerless: <owner> — <reason>` beside the code, which cannot be bulk-regenerated (the pgw#1425 failure mode).

It found two more siblings within the hour:

| sibling | disposition |
|---|---|
| `_set_execution_lane` | **WIRED** — every v2 request reported the property's `bf16-w16a16+eager` default whatever it ran, and a `handles=[...]` body branched on it |
| `_set_config` | **MARKED with its ruling** — `@entrypoint` declares no config, so there is no schema to decode `RunJob.config_params` against. A design call, not a repair; `subproc.py:194` still reads `_config_snapshot`, so it is not dead either |

## Also

* a reserved-field `ValidationError` is now **INVALID**, not FATAL — retrying an unparseable ref costs a pod-minute to reach the same refusal;
* `serving/entrypoints.py:347-349` no longer attributes this symptom solely to a missing `kind=`. That note sent the first reader hunting for a declaration that was already right.

## Verification

`tests/test_reserved_repo_wires_pgw1475.py` — **7 passed**. A real `@entrypoint(kind="conversion")` body keeping `source_from_ctx`'s raise site verbatim, over a REAL projected snapshot published by the production chokepoint (`ensure_snapshot_async`) into a real CAS, reached through the real `TENSORHUB_CACHE_DIR` loader — no patched seam. The **RED-PROOF** stubs the added call back out and gets the measured message character for character. The `clone-huggingface`-shaped arm passes **with the wire and without it**, which is what makes the file measure the defect rather than the fixture.

No regressions: `test_serve_path_wires_pgw1425` + `test_serve_loop_envelope` + `test_endpoint_discovery` + `test_weightless_entrypoints` + this suite = **48 passed, 2 failed**, and those 2 are exactly master's standing pair (`…_with_fake_tensors`, and `tensorfs.contracts` pending tensorfs#111). `ruff` clean. `mypy` shows master's own two rows in the locked env, neither in a file this diff touches. The `lint_unreached_surface` gate carries 4 pre-existing `trace_context` rows from master (which had 5) — none of them this lane's.

### DIFF
--- a/scripts/lint_unreached_surface.py
+++ b/scripts/lint_unreached_surface.py
@@ -1226,6 +1226,86 @@ def gate_rows_without_reasons(
         if is_gate_label(label) and not reasons.get(label))
 
 
+# --------------------------------------------------------------------------
+# pgw#1475: WRITER-LESS PRIVATE SETTERS — the blind spot that cost 25 producers
+# --------------------------------------------------------------------------
+#
+# `collect_definitions` SKIPS every `_`-prefixed name (a private callable is
+# not surface), which is right for the surface question and exactly wrong for
+# one shape: a private SETTER whose public reader is the contract. Nothing on
+# the ratchet ever mentioned `RequestContext._set_source_path`, because it was
+# never a candidate — so when pgw#1373 deleted its only caller with
+# `executor.py`, `ctx.source_path` kept returning `None` forever and 25 of 27
+# conversion producers died on their own first line at 0 GPU-seconds.
+#
+# A reader with no writer renders its failure as ABSENCE, which is why no test
+# caught it: the property exists and answers `None`. So this check has NO
+# BASELINE FILE, deliberately — bulk-appending rows to one is what made a
+# correct instrument silent in pgw#1425. Its only escape valve is an INLINE
+# `# writerless: <owner/issue> — <reason>` comment on the `def` line or the
+# line above it: one sentence, next to the code, un-regenerable in bulk, and
+# it REFUSES an empty reason.
+
+#: `_set_<x>` on a context class is the WRITE half of the public `<x>` reader.
+_SETTER_RE = re.compile(r"^_set_[a-z0-9_]+$")
+_WRITERLESS_OK = re.compile(r"#\s*writerless:\s*(?P<reason>\S.*)$")
+
+
+def _writerless_marker(lines: List[str], lineno: int) -> Optional[str]:
+    """The inline sentence excusing a writer-less setter, or None."""
+    for idx in (lineno - 1, lineno - 2):
+        if 0 <= idx < len(lines):
+            m = _WRITERLESS_OK.search(lines[idx])
+            if m:
+                return m.group("reason").strip()
+    return None
+
+
+def writerless_private_setters() -> List[Tuple[str, Path, int]]:
+    """Private `_set_*` methods defined in src/ that nothing in src/ calls."""
+    files = py_files(POD_ROOTS)
+    defined: Dict[str, Tuple[Path, int]] = {}
+    excused: Set[str] = set()
+    called: Set[str] = set()
+    for path in files:
+        try:
+            text = path.read_text(encoding="utf-8")
+            tree = ast.parse(text, filename=str(path))
+        except SyntaxError:
+            continue
+        lines = text.splitlines()
+        for node in ast.walk(tree):
+            if (isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
+                    and _SETTER_RE.match(node.name)):
+                defined.setdefault(node.name, (path, node.lineno))
+                if _writerless_marker(lines, node.lineno):
+                    excused.add(node.name)
+            elif isinstance(node, ast.Call):
+                fn = node.func
+                # `ctx._set_source_path(p)` and `(set_path or ctx._set_x)(p)`
+                # are both calls; a bare `getattr(ctx, "_set_x")` reference is
+                # covered by the Attribute/Constant walk below.
+                if isinstance(fn, ast.Attribute) and _SETTER_RE.match(fn.attr):
+                    called.add(fn.attr)
+                elif isinstance(fn, ast.Name) and _SETTER_RE.match(fn.id):
+                    called.add(fn.id)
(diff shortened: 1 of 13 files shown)


PaulFidika:
**Merging with `--admin`, recorded** (the 2026-08-19 skip-CI ruling — `--admin` is the DEFAULT during this campaign). Both reds are MASTER'S, verified by comparison rather than assumed:

**`fast gates`** — mypy `Success: no issues found in 575 source files`, ruff `All checks passed!`, `lint_unreached_surface` **exit 0** (and its siblings-selftest green), `lint_pickle_readers` / `lint_test_fork_start_method` / `lint_incident_test_names` selftests all green. The one failing step is `lint_incident_test_names`: **8 incident-named test modules**, every one another lane's — pgw#1421 (×2), #1425, #1457, #1461, #1468, #1471, #1479. This lane's own suite was RENAMED to `tests/test_reserved_repo_contract.py` rather than added to that baseline (`3d18403a`), so the count went 9 → 8 on this branch.

**`tests`** — 2 failures, both `tests/test_weightless_program_load_pgw1468.py` (`test_a_STOCK_load_of_a_real_weightless_blob_FAILS`, `test_a_UNIFORM_shape_blob_hides_the_defect_which_is_why_it_survived`), a file this diff does not touch. **They are FLAKY on the CI runner, not broken by this PR**: run [32250937931](https://github.com/cozy-creator/python-gen-worker/actions/runs/32250937931) — this lane's code, one commit earlier — was **`tests` SUCCESS, 2110 passed**. They pass locally too. Master itself has not run the suite since pgw#1468 landed, because `uv sync --locked` refused (see below).

**Three master-owned reds this lane cleared on the way**, each in its own commit so they are separable:
* `fef4aa9f` — `uv.lock` caught up to pgw#1478's torchcg rev. pgw#1477 fixed this exact drift six hours earlier; `uv sync --locked` is the FIRST step of every job, so **no PR in this repo was getting any signal at all**. This PR is the first to reach the suite since.
* `9abd8a4f` — pgw#1479's `_publish` helper got its return annotation, the only mypy error between master and a green type check.
* `da3ee01e` — pgw#1479 also broke `test_producer_declarations`'s hand-built `ServeLoop`… no: that one WAS this lane's, and it is fixed by STATING the new field rather than defaulting it, because `getattr(self, "_hf_token", "")` in production would turn "this loop was never given the pod's HF credential" into "the pod has none".

PaulFidika:
Rebased onto `dcea84d7`. Two of the three master-owned fixes above were **dropped as already upstream** (pgw#1477 landed the same `uv.lock` catch-up, and pgw#1479's annotation landed too) — the rebase says so itself: `patch contents already upstream`. The `worker.py` conflict with pgw#1474 was resolved by KEEPING BOTH: its traceback-tail comment and arm, plus this lane's `ApiValidationError` in the same tuple.

Re-verified on the rebased tree with the locked env: mypy **`Success: no issues found in 584 source files`**, ruff **`All checks passed!`**, `lint_unreached_surface` **exit 0**, and `tests/test_reserved_repo_contract.py` + `test_producer_declarations.py` + `test_serve_path_wires_pgw1425.py` = **43 passed**.

`lint_incident_test_names` is now at **13** rows on master (up from 8 in the hours this PR was open, as other lanes landed) and **none of them is this lane's** — the suite here is `test_reserved_repo_contract.py`. That gate is the sole remaining `fast gates` step that fails, and it fails identically on master.

PR #1044, opened 2026-08-19
pgw#1477 RED PROOF — DO NOT MERGE: the pgw#1460 divergence, replayed

PaulFidika:
Throwaway. Recreates `fad60a8b`'s exact shape on top of the pgw#1477 rail: `pyproject.toml`'s torchcg pin moved to `12c99841`, `uv.lock` left at `6c91b83d`, no re-lock.

Acceptance evidence for pgw#1477: `fast gates` must fail at its FIRST step (`pgw#1477 guard`), alone, naming both files and both revs — instead of three jobs dying at `uv sync --locked` with a message that names neither.

Closed unmerged once observed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -188,6 +188,24 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # pgw#1477 — FIRST, ahead of uv itself, and stdlib-only so it can be.
+      # `uv sync --locked` (two steps down, and the first step of every other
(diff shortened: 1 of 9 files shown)


PaulFidika:
Red proof observed: `fast gates` failed at its FIRST step (`pgw#1477 guard`) in 7s — the only failing step — naming both files and both revs, and the `--selftest` arm passed in the same run. Same in `drift` (8s) and `tests` (5s). Job 96057183687. Closing unmerged; the rail is PR #1045.

PR #1045, opened 2026-08-19
pgw#1477: the torchcg pin had three spellings and nothing made them agree — the re-vendor is ONE act, and a stale lock is ONE self-describing red

PaulFidika:
Closes pgw#1477.

`uv sync --locked` is the FIRST step of EVERY job in EVERY workflow. A re-vendor that moved
`pyproject.toml`'s `[tool.uv.sources]` torchcg rev and left `uv.lock` behind therefore killed
`fast gates`, `tests` AND `drift` at install, on PRs that had touched neither file, with a message
naming no file and no rev. Twice on 2026-08-19, twenty minutes apart — and nothing in the required
set could ever have caught it: a gate cannot fail on something it never got far enough to read.

## Which writer each breakage used

Both, the same one: a HAND EDIT of the pin inside a re-vendor. There was no vendor script to fix —
`_vendor/VENDORED.toml`'s prose said `git archive <rev> <subdir> | tar -x`, re-apply the rewrites,
regenerate the digests, and the pin moves were three separate manual edits of which `uv lock` was
the forgettable one. `fad60a8b` (pgw#1460/tcg#58) and `3966f323` (pgw#1478/tcg#61) each touched
`pyproject.toml` + `VENDORED.toml` + the vendored tree and neither touched `uv.lock`.

## Half 1 — the writer re-locks

`scripts/vendor_snapshot.py` (`task vendor -- <package> <rev>`) is now the ONE WRITER of a vendored
rev, and it re-locks in the same act: resolve the rev, extract the subdir, re-apply the declared
mechanical rewrites, regenerate the sha256 table, write the rev into `_vendor/VENDORED.toml` AND
`pyproject.toml`, run `uv lock`, stage all four paths, then re-run the checker on its own output.
Packages whose rewrites are NOT mechanical (tensorfs, whose `planner.py`/`contract.py` are
hand-written ports of deleted upstream modules) are refused the file copy by name and re-pinned
with `--pins-only`, which does everything else.

Proven by a **no-op re-vendor at the rev already vendored**: 39/39 files byte-identical, digest
table unchanged, `uv lock` a no-op. The only diff is `VENDORED.toml`'s rev normalised from 8 hex to
40 — a prefix is not comparable to what uv writes into `uv.lock`.

## Half 2 — the check runs FIRST and fails ALONE

`scripts/lint_lock_pin_agreement.py` runs ahead of uv itself in `fast gates`, `tests`, `drift` and
`publish`, and as the first line of `task lint` (every other line there is a `uv run`, which
silently RE-RESOLVES a stale lock instead of refusing it — so `task lint` was green on exactly the
tree CI dies installing). Stdlib only, no venv, no network, ~30 ms. It reads all THREE spellings of
the rev — `[tool.uv.sources]`, every git URL in `uv.lock`, and the vendored snapshot's — and names
both files and both revs:

```
  - torchcg: the pin in pyproject.toml and uv.lock DISAGREE.
      pyproject.toml  [tool.uv.sources].torchcg  rev = 12c99841f01f9ccba22c279b72bebc30bd7b22d2
      uv.lock         https://github.com/cozy-creator/torchcg  rev = 8e8c5ca6acdb3ed13f94ae6878c8248114ecd9c7
```

`--selftest` runs in `fast gates` and is what makes the green a result: it builds a divergent tree
in a tempdir and fails if the check stays quiet on it.

## Red / green on the REAL divergence

Replayed against actual history (`pyproject.toml` + `uv.lock` + `VENDORED.toml` at each sha):

| commit | verdict |
|---|---|
| `fad60a8b` (occurrence 1) | **RED** — pin `12c99841` vs lock `8e8c5ca6` |
| `3966f323` (occurrence 2) | **RED** — pin `6c91b83d` vs lock `8e8c5ca6` |
| `c16ebf928` ("fixed" occ. 1) | **RED** — its tree was already stale against the re-vendor that landed 7 minutes earlier |
| `bdecee16` (occurrence 2's fix) | green |
| current master `3c123333` | green — **no third breakage; master is clean through the rail** |

Plus a live CI red-proof on a throwaway branch (PR #1044, closed unmerged): `fad60a8b`'s exact
shape on top of this rail, and `fast gates` dies at its FIRST step naming both files instead of
three jobs dying at `uv sync --locked`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -188,6 +188,24 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # pgw#1477 — FIRST, ahead of uv itself, and stdlib-only so it can be.
+      # `uv sync --locked` (two steps down, and the first step of every other
+      # job in every other workflow) dies on a stale lock with a message that
+      # names neither file and no rev — so a torchcg re-vendor that moved
+      # pyproject.toml's pin and not uv.lock turned `fast gates`, `tests` and
+      # `drift` red simultaneously, at install, on PRs that had touched
+      # neither file. Twice on 2026-08-19, twenty minutes apart. Nothing in the
+      # required set could report it: a gate cannot fail on something it never
+      # got far enough to read.
+      #
+      # ~30 ms, no venv, no network. The `--selftest` arm runs here only, and
+      # is the reason this line is evidence: it builds a divergent tree in a
+      # tempdir and fails if the check stays green on it.
+      - name: pgw#1477 guard - uv.lock agrees with pyproject.toml's git pins
+        run: |
+          python3 scripts/lint_lock_pin_agreement.py --selftest
+          python3 scripts/lint_lock_pin_agreement.py
+
       # pgw#1264: `@v6` HERE ONLY, and the version is the point. `@v4` bundles
       # `@actions/cache` v3, which speaks the cache API GitHub retired, so
       # `enable-cache: true` on `@v4` is dead config that reads as a working
@@ -655,6 +673,11 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
+
+      # pgw#1477: see `fast gates`. Same guard, same reason — this job's
+      # `uv sync --locked` is one of the three reds a stale lock produces.
+      - name: pgw#1477 guard - uv.lock agrees with pyproject.toml's git pins
+        run: python3 scripts/lint_lock_pin_agreement.py
 
       - name: Install uv
         uses: astral-sh/setup-uv@v4
--- a/.github/workflows/proto-contract.yaml
+++ b/.github/workflows/proto-contract.yaml
@@ -57,6 +57,12 @@ jobs:
       - name: proto drift check (vs PROTO_DIGEST)
         run: PROTO_SKIP_PEER=1 scripts/proto-drift-check.sh
 
+      # pgw#1477, same layer-1 argument: `uv sync --locked` below is where a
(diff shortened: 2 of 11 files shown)


PaulFidika:
**`--admin` merge, recorded.** `fast gates` (the ruleset's only required context) is red on MASTER's own standing set, not on anything in this diff — which is the exact failure mode this PR exists to make legible.

* **mypy** — was red on master from three missing annotations landed by `b8644cbc` (pgw#1479) and pgw#1481. **Fixed here** (`8f530556`); mypy is green in run 32251061884.
* **`pgw#1362 guard - a NEW test module is not named for an issue`** — 13 incident-named modules, **every one of them already on `origin/master`** (verified with `git cat-file -e origin/master:tests/<f>.py` for all 13) and **none added by this PR** (`git diff --stat origin/master...HEAD` adds no test module). It was invisible until now only because it sits AFTER the mypy step, which was failing first. Its fix is a `git mv` per module, owned by pgw#1362's burn-down and by the lanes that landed them — not by this diff.
* **`tests`** — PASSED on the identical tree one push earlier (run 32250274238, 7m35s). The run here aborted with the pgw#1316 `Event loop is closed` / `Fatal Python error: Aborted` shape, which this repo documents as "a suite failure nobody caused". `tests` gates nothing (pgw#1264).
* **`drift`** — pass, 49s.
* **This PR's own guard** — `pgw#1477 guard - uv.lock agrees with pyproject.toml's git pins`: **success**, in every run, including its `--selftest` arm.

PR #1047, opened 2026-08-19
pgw#1362: the intake guard's own 13 refusals — every incident-named module on master takes its DOMAIN name

PaulFidika:
`fast gates` is the master ruleset's **only** required context, and it has been red
on every lane: thirteen test modules on `origin/master` carry an issue id in the
filename, and the pgw#1362 intake guard refuses each one by name. The guard is
right, and the fix it prints is `git mv`. This is that, and nothing else.

| old | new |
|---|---|
| `test_adopt_refusal_permanence_pgw1472.py` | `test_adopt_refusal_permanence.py` |
| `test_engine_runtime_pgw1421.py` | `test_engine_runtime.py` |
| `test_engine_runtime_binding_pgw1421.py` | `test_engine_runtime_binding.py` |
| `test_env_identity_pgw1472.py` | `test_env_identity.py` |
| `test_mint_child_end_to_end_pgw1457.py` | `test_mint_child_compile.py` |
| `test_mint_publish_shape_pgw1471.py` | `test_mint_publish_shape.py` |
| `test_modular_model_index_pgw1481.py` | `test_modular_model_index.py` |
| `test_phase_enum_emitters_pgw1480.py` | `test_phase_enum_emitters.py` |
| `test_scratch_repo_derives_its_release_pgw1479.py` | `test_scratch_repo_release.py` |
| `test_serve_path_wires_pgw1425.py` | `test_serve_path_wires.py` |
| `test_trace_context_surface_pgw1461.py` | `test_trace_context_surface.py` |
| `test_variant_bridge_pgw1473.py` | `test_variant_bridge.py` |
| `test_weightless_program_load_pgw1468.py` | `test_weightless_program_load.py` |

The two engine modules keep their split rather than collapsing onto one name:
one drives process supervision in isolation against a stand-in binary, the other
binds the declaration to the runtime through discovery and the host. Same
subject, different boundary.

**The baseline is untouched, deliberately.** All thirteen were flagged as NEW,
not grandfathered — there is no row to delete in `scripts/incident_named_tests.txt`
and no `BASELINE_HIGH_WATER` to lower. It stands at 126.

**Content diff is two lines**, both cross-references a rename forces:
`trace_context.py`'s docstring names its fence, and the binding module's
docstring names its isolated sibling. No test body changed.

### Evidence

- `scripts/lint_incident_test_names.py`: **13 problems -> 0**; `--selftest` 13 cases OK.
- Collection over `tests/`: **2183 before, 2183 after** (the single `jsonschema`
  collection error is pre-existing on master and unrelated).
- The eight light renamed modules RUN: 73 passed.
- `mypy src/gen_worker tests` clean, 581 files; `ruff check src/gen_worker` clean;
  mypy ratchet unchanged at 339.

### One note for two open lanes

#1040 (`1475-reserved-repos`) carries a 2-line edit to `..._pgw1479.py` and #1022
(`1462-derive-vendored`) a 7-line edit to `..._pgw1461.py`. Both deltas are tiny
against 174/225-line files, so git's rename detection carries them onto the new
name on merge — but both lanes should rebase rather than resolve by hand.


PaulFidika:
### `fast gates` is GREEN — and clearing this red uncovered a second one

Run [32253319355](https://github.com/cozy-creator/python-gen-worker/actions/runs/32253319355): **fast gates: success**, zero failed steps. The pgw#1362 guard step passed on the run before it too, which is how the second red became visible.

**`fast gates` stops at its first failing step, so master has been carrying a STACK of reds where only the topmost is ever visible.** Step 7 (mypy) was red until pgw#1477; step 17 (incident names) was red until commit 1 here; **step 27 (pgw#1215, no export/AOTI compile in the serving process) has been red the whole time and nothing could report it.** A lane reading `fast gates: failure` learns the count is at least one and nothing more.

So this PR carries a second commit. `_vendor/torchcg/hollow.py:190` calls `torch.export.load` — the device-checked reader for an archive the derive itself just wrote. `hollow` is the weights-free derive session (tcg#45) and sits *behind* two modules already declared `CHILD_ONLY` rather than beside them: `_vendor.torchcg.discovery` (`from .hollow import HollowSession`) and `release.derive` (`importlib.import_module("torchcg.hollow")`). Those are its only two importers in the tree, `_vendor/torchcg/__init__.py` deliberately withholds `hollow` from the package root, and no serving-package module imports any of the three. The row exempts `hollow`, not its importers.

**Red-verified**, because an exemption that cannot fail is a hole: stub the call to `return None` and the fence refuses the row by name (*"CHILD_ONLY names `_vendor.torchcg.hollow`, which no longer calls any banned symbol"*).

Also run locally against this tree: all 33 `fast gates` guard invocations plus the 8 digest/changelog gates, **0 red**.

PR #1055, opened 2026-08-19
pgw#1489: the artifact key is (trace digest, sm, compile stack) — the installed-set closure is deleted

PaulFidika:
Paul, direct P0 (DESIGN-RULINGS 2026-08-19 addendum 4 as corrected + the full-artifact-axis entry). The env half of an artifact key is the COMPILE STACK — torch, triton, `nvidia-*` — read off the endpoint's own `uv.lock`. Everything else is locked by that lock and cannot reach the compiler.

**Deleted, not deprecated:** `closure_hash` over the whole resolved package set, and `installed_closure()` restating it from `importlib.metadata`. Two representations of one environment that could never agree (pgw#1472 measured all three reasons), splitting the artifact pool on drift no compiler can see.

- `env_identity` is the ONE definition site: the derive stamps from it and the serve runner states from it, so a locally derived document is adoptable by the local serve by construction.
- **Flavored locks** (one `uv.lock`, several CUDA buckets via conflicting extras) resolve PER BUCKET. The walk is breadth-first from the bucket's own torch pin, because a shared package (cudnn is locked once) carries an edge to the OTHER bucket's cublas and a depth-first read silently keys cu126 artifacts with cu128 libraries. Verified against a real `uv lock` of a two-extra project. The host contributes only (sm, bucket) and resolves nothing; the bucket is not a separate key field because torch's own version states it.
- Adoption: `stack=` replaces `installed=` end to end. The venv is compared to the stated stack DIAGNOSTICALLY (a warning naming the package), never as a gate.
- A derive with no lockfile refuses by name. Dead `installed` plumbing on BackgroundMint/SelfMint (threaded, never read) is gone.
- The hub answer must state `env_compile_stack`; the old `env_lockfile_hash` shape is refused by name (tensorhub half is landing as a matched pair).
- `scripts/lint_no_installed_set_keying.py` (gates + CI) keeps the installed set out of the key path, with one allowlisted diagnostic. Proved red on a reintroduced enumeration, green after.

## Red/green by EXECUTION, on tonight's real 14 sd1.5 artifacts, one irrelevant 4-package diff

| | RED (master, closure key) | GREEN (this branch) |
|---|---|---|
| same key across the diff | no | **yes** |
| holes under my env | 0/14 | 0/14 |
| holes under the drifted env | **14/14** | **0/14** |
| adopt audit under the drift | REFUSES | passes |
| `sm_120` | — | refuses BY NAME, 14/14 holes |
| `torch 2.14.0` | — | 14/14 holes |

**Rekey ledger:** 32 artifacts + 32 manifests re-positioned in the local graph CAS (sd15 14, sdxl 18), **0 re-mints** — content-addressed, so only refs moved. sd15 and sdxl had different closures and now share one env key.

torchcg PR #59 merged as `3d958bf9` and vendored here (suite: 434 passed, 10 skipped). mypy + ruff clean.


PaulFidika:
## ACCEPTANCE, on the card (4070 Laptop, sm_89) — 14/14 adopted, byte-identical image

The pgw#1460 sd15 serve harness, pointed at this branch and at the rekeyed local graph CAS, with `stack=` read from sd15's own `uv.lock` (`bucket=cu130`):

```
adopted        14 / 14      holes 0
armed_on_unet  14           dispatcher _ForwardDispatcher
boot           59.6 s
run 0          5.18 s       compiled_graph_calls 26   (25 steps x 1 unet call + 1)
peak           3199 MiB allocated / 3946 MiB reserved
```

**Image sha256 `aced848a991dbf4efa677228225d7a92966bdfc935f369fb5e5bdae337b61fc4`** — identical to the pgw#1460 run under the OLD key, both of its runs. The rekey moved positions and nothing else: same artifacts, same bytes out.

No Xid/NVRM events during the window; card back to 198 MiB.

Matched pair: tensorhub#1536 (the hub answers `env_compile_stack`; both DB integration suites green on the new schema).

PR #1091, opened 2026-08-20
pgw#1525 (store half): re-vendor tcg#69 — the graph store heals itself, and the remedy is printed for the MISS

PaulFidika:
Consumer half of **[tcg#69](https://github.com/cozy-creator/torchcg/pull/66)** (torchcg `f45fc64a`, merged, CI green). `LocalGraphStore.get_graphs` now answers a clean **MISS** on a document this build cannot decode, discards the stale bytes, and says so once. Re-vendored as one act (`task vendor`) so `pyproject.toml`, `uv.lock` and `VENDORED.toml` name the same rev.

### Why upstream and not the vendor copy

`tests/test_vendored_snapshot_pgw1310.py` fences the snapshot per-file sha256 — a hand-patched vendor tree is refused by construction. And this repo runs **two** torchcg copies on purpose: the snapshot serves the mint and the serving path, the `dev` git pin serves the derive. A vendor-only fix leaves the derive on the old behaviour, with the failure landing two tools from the cause — the wound `VENDORED.toml` documents.

### Relationship to #1090 — which is merged and is NOT superseded

#1090 is the CLI half, and its own commit message names the split: *"The STORE side — answering a clean miss and GCing the unreadable leftover — is torchcg's and is deliberately not in this diff."* This is that side. #1090 fixed **one** call site; `serving/serve_adoption.py`, `serving/__main__.py` and `mint_store.py` all call `get_graphs` bare and were still exposed. Fixing the store fixes them together.

### The part that is not mechanical

Once the local store stops raising, #1090's `except StoreError` branch stops firing for it — **and the operator-facing lines that named the remedy went with it.** Printing them only on a raise would silently retire the message on the exact path pgw#1525 exists to fix. So the remedy is printed for the MISS itself:

```
adopt: no compiled-graph document for <module> in <store>.
adopt: SERVING EAGER. Compiled graphs are derived and disposable, so this
       costs speed, never correctness.
adopt: remedy — `gen-worker compile` mints this endpoint's graphs for this
       card in the current format.
```

Saying one thing for a stale store and an empty one is not lost information — it is #1090's own conclusion that the two are the same fact. WHICH it was is still stated by torchcg's discard warning, naming the graph-set, the bytes and the decode failure. **Verified rather than assumed** that this reaches a terminal with no logging configured (`logging.lastResort` handles WARNING+).

The `except StoreError` branch **stays** — the hub-backed store is a different implementation of the same protocol and still raises. It now has a test driving it through a store double, so the call site is covered independently of the backend.

### Tests

The assertion that could no longer be true (`graph_store_unreadable` on the local path) is **replaced, not deleted**: one test asserts the discard IS announced with what / which bytes / why, and one plants **the shape actually found on disk**. That shape is `v=2` with a *valid* field set; the existing fixture is `{"v": 1, ...}` and fails on the FIELD SET, so it would have passed a reader that only rejected unknown fields.

22 passed in the touched suites · 37 across the adopt-adjacent suites · `mypy` clean over 633 files · `ruff check src/gen_worker` clean · `task lock-check` green.

Field evidence (pgw#1525's own box): `~/.cache/cozy/graph-cas` held `sd15.main` + `sdxl.main` at `v=2` → `StoreError: document v must be 3` at adopt time. 1.08 GiB stranded, 144 of 193 objects unreachable from any ref.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -696,7 +696,7 @@ torchvision = [
 # uv.lock behind, which breaks `uv sync --locked` — the first step of every CI
 # job — for every open PR; `scripts/lint_lock_pin_agreement.py` now refuses that
 # tree locally and as CI's first step.
-torchcg = { git = "https://github.com/cozy-creator/torchcg", rev = "9fc1f2ac93a1a3976090c77ca27297e68dbfa88e" }
+torchcg = { git = "https://github.com/cozy-creator/torchcg", rev = "f45fc64a9ba1323e0a44975565d822517aab8058" }
 # Saying which pins were deleted requires naming them.
 # retired-name: historical rationale, not a live reference.
 # pgw#1310 deleted the `hashrepo` / `torch-compiled-graphs` git pins that stood
--- a/src/gen_worker/_vendor/VENDORED.toml
+++ b/src/gen_worker/_vendor/VENDORED.toml
@@ -306,7 +306,7 @@ rewrites = [
 
 [packages.torchcg]
 repo = "https://github.com/cozy-creator/torchcg"
-rev = "9fc1f2ac93a1a3976090c77ca27297e68dbfa88e"
+rev = "f45fc64a9ba1323e0a44975565d822517aab8058"
 subdir = "src/torchcg"
 # pgw#1474 / tcg#61 RE-VENDOR (2026-08-19), from `6c91b83d`. `CompiledGraphCall`
 # stops pre-flattening the call: the loaded package is torch's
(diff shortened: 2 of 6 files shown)


PaulFidika:
## `tests` is red for exactly 2 tests, and both are red on `master` — proof, not assertion

CI run [96288540612](https://github.com/cozy-creator/python-gen-worker/actions/runs/32322956133/job/96288540612):

```
===== 2 failed, 2430 passed, 47 skipped, 62 warnings in 356.16s (0:05:56) ======
FAILED tests/convert/test_flavor_token_deleted_pgw1319.py::test_the_classifier_is_deleted_and_the_vocabulary_survives
FAILED tests/convert/test_precision_class_stamp_pgw1300.py::test_the_ladder_module_no_longer_exports_a_placement
```

The same two, run on a clean canonical `master` checkout at `1edc5ecb` (this PR's base):

```
$ uv run --extra dev pytest tests/convert/test_flavor_token_deleted_pgw1319.py \
                            tests/convert/test_precision_class_stamp_pgw1300.py -q
FAILED tests/convert/test_flavor_token_deleted_pgw1319.py::test_the_classifier_is_deleted_and_the_vocabulary_survives
FAILED tests/convert/test_precision_class_stamp_pgw1300.py::test_the_ladder_module_no_longer_exports_a_placement
2 failed, 39 passed in 23.00s
```

Both assert `convert/ladder`'s precision vocabulary (`svdq-fp4` vs `svdq-int4` in `PRECISION_CLASSES` / `__all__`). **This PR touches no file in `src/gen_worker/convert/`** — the diff is `_vendor/torchcg/store.py` (re-vendored), `VENDORED.toml`, `pyproject.toml`, `uv.lock`, `cli/daemon.py`, `tests/test_cli_adopt_degradation.py`.

The cardinality is the load-bearing part: 2 failed on the PR, the same 2 on master, so this branch adds **zero** new failures across 2430 passing tests. Not absorbing an unowned red (pgw#1521's standard).

`fast gates` ✅ · `drift` ✅ · locally: `mypy` clean over 633 files, `ruff check src/gen_worker` clean, `task lock-check` green.

The `convert/ladder` vocabulary drift needs its own owner and is unrelated to pgw#1525.

PR #1098, opened 2026-08-20
pgw#1525: `compile`'s re-derive branch was UNREACHABLE — both spellings of a missing program are a miss

PaulFidika:
`cli/compile.py`'s module docstring calls the re-derive load-bearing: *"when this machine's graph CAS does not hold the program for a specialization, `compile` RE-DERIVES it locally (~2 min CPU)."* **On the one state that sentence is about — a box that has never derived this endpoint — the branch could not execute.**

### Root cause: a miss rendered as an exception (tcg#69's defect class, one layer up)

Two implementations of one `GraphStore` protocol disagree about what a miss IS:

| store | `fetch_program` on a miss |
|---|---|
| `torchcg.store.LocalGraphStore` | returns `None` |
| `serving.mint_store.WorkerGraphStore` | **raises `ProgramBlobUnreachable`** — its own message calling the condition *"ORDINARY"* |

`_store()` builds the second; `_ensure_program` tested only for `None`, so the raise escaped past `rederive()`.

**Measured** on a wiped graph-cas, canonical master, the real verb with no flags:

```
gen-worker compile: failed=14 (of 14)      exit 1, 6.04 s
```

— having never logged `re-deriving`, and every failure telling the user the remedy was to run `compile`. A verb whose job is a half-hour of compilation returning in **six seconds** is what a dead branch looks like.

### Why nobody saw it

pgw#1491's acceptance ran `compile` against a **full** store and recorded `present=14 of 14`. Every specialization short-circuits at `store.has_artifact()` **before `_ensure_program` is called**; the other arm (`--vram-budget 1.0` → `below_floor=14`) returns before the loop. **`compile`'s build path has never been executed by anything.** An acceptance run on a warm box cannot certify a verb whose whole job is the cold box.

### The fix

Both spellings of a miss are a miss, and the answer to a miss is to derive. The `rotten` arm of the same refusal is deliberately included — a serialized program is a derived, disposable, machine-scoped artifact, so bytes failing their own scrub deserve what absent bytes deserve. Two refusals survive because they are load-bearing: a **malformed graph key** is checked by `is_graph_hash` up front (by the predicate, never by matching the refusal's wording — that would leave this one prose edit from paying a two-minute derive per specialization), and a program still missing **after** the derive stays the typed `CompileError`.

### Necessary, not sufficient — and the tracker says so

With this in, the same run gets further and hits three more defects, all filed under pgw#1525:
1. the re-derive passes the endpoint **source** dir as `checkpoint_dir`, and `compile` has none of `lock`'s `--checkpoint` flags because it declares itself WEIGHTLESS — both cannot be true for a model-bearing endpoint (**needs a Paul-level ruling; filed, not chosen**);
2. the mint child hard-codes `/usr/local/cuda`;
3. worst — a completed `built=14 (of 14)` writes **every artifact to `torchcg/v1/graphs`**, the band pgw#1373 retired, so adoption sees **0 artifacts / 14 holes after 26 minutes of real GPU work.**

### What it does buy, measured tonight on a 4070

With this fix the mint runs end to end: **derive 127.2 s + build 1561.1 s = 28.1 min** for sd15's 14 specializations — va#3's *"paid once ever"* column, previously ABSENT.

### Red arm

4 mutations, each applied 5x, **all CAUGHT 5/5**: the pre-fix `raise` · a swallowed miss that never derives · a malformed key treated as a miss · a present program made to pay for a derive. 6 tests, parametrized over **both** store dialects, because a fix written against one store is how this happened.

`ruff check src/gen_worker` clean · `mypy` clean on both changed files.

### DIFF
--- a/src/gen_worker/cli/compile.py
+++ b/src/gen_worker/cli/compile.py
@@ -241,22 +241,65 @@ def _ensure_program(spec: Spec, store: Any, rederive: Any, scratch: Path) -> Pat
     re-deriving locally — never by fetching, because there is nothing portable
     to fetch. A malformed key is a typed refusal from the store rather than a
     miss, so a wiring bug can never spend a pointless two-minute derive.
+
+    A MISS ARRIVES TWO WAYS, AND THIS ASKED FOR ONLY ONE (pgw#1525).
+    ``LocalGraphStore.fetch_program`` returns ``None`` for a miss;
+    ``WorkerGraphStore.fetch_program`` — which is what ``_store`` actually
+    builds — RAISES ``ProgramBlobUnreachable``, and its own message calls the
+    condition "ORDINARY". This function only tested for ``None``, so on the one
+    state it exists for — a box that has never derived this endpoint — the
+    raise escaped past ``rederive()`` and the re-derive branch was unreachable.
+    Measured: `compile` on a wiped store answered ``failed=14 (of 14)`` in
+    6.04 s having never logged "re-deriving", and told the user to run
+    `compile`.
+
+    Both spellings are now the same miss, and the answer to a miss is to derive.
+    That includes the ``rotten`` arm of the same refusal: a serialized program
+    is a DERIVED, DISPOSABLE, machine-scoped artifact, so bytes that fail their
+    own scrub deserve exactly what absent bytes deserve — regeneration, not a
+    report. The refusal survives where it is load-bearing: a malformed graph key
+    is not a miss, so it is re-raised rather than paying for a derive that
+    cannot possibly satisfy it, and a program still missing AFTER the derive is
+    the typed ``CompileError`` below.
     """
+    from .._vendor.torchcg import is_graph_hash
+    from ..serving.mint_store import ProgramBlobUnreachable
+
+    # Checked HERE, by the predicate, rather than recognised from the refusal's
+    # wording. A malformed key is a wiring bug and no derive can satisfy it, so
+    # it must not be swallowed as a miss; matching on message text to tell the
+    # two apart would put this function one prose edit from silently paying for
+    # a two-minute derive on every specialization.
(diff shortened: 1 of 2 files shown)


PaulFidika:
## `tests` is red for 2 tests; the same 2 are red on master — proof, not assertion

This PR ([run](https://github.com/cozy-creator/python-gen-worker/actions/runs/32329504031/job/96307301965)):

```
===== 2 failed, 2493 passed, 48 skipped, 91 warnings in 452.93s =====
tests/test_image_encoding_default.py::test_webp_lossless_is_reachable_and_pixel_exact
tests/test_mint_child_compile.py::test_the_mint_child_compiles_a_real_exported_program
tests/test_mint_child_compile.py::test_the_child_request_carries_no_graph_interface_to_get_wrong
```

**PR #1098 (a different lane, `1530-lane-dtype-precedence`, since MERGED)** — [run](https://github.com/cozy-creator/python-gen-worker/actions/runs/32329375064/job/96306935190), same era, same job:

```
===== 2 failed, 2489 passed, 48 skipped, 93 warnings in 424.45s =====
```

…failing in the **same two files**. Same cardinality, same tests, and that PR was merged.

This PR's diff is **`src/gen_worker/cli/compile.py` + one new test file**. It touches nothing in image encoding and nothing in the mint child's compile path. `test_image_encoding_default.py` passes locally on a clean canonical `master` checkout (`14 passed`), so the webp arm looks environment-bound to the runner (Pillow/libwebp), not code-bound — either way it is not this branch.

`fast gates` ✅ · locally: `ruff check src/gen_worker` clean, `mypy` clean on both changed files, 6/6 new tests pass, and the red arm is **4 mutations × 5, all CAUGHT 5/5**.

Not absorbing an unowned red (pgw#1521's standard).

PR #1099, opened 2026-08-20
pgw#1533: `compile` succeeds only when the SERVING READER can find what it wrote

PaulFidika:
`gen-worker compile` reported `built=14 (of 14)`, exit 0, after 26 minutes of real inductor work on a 4070 — and boot-time adoption then saw **0 artifacts and 14 holes**. Every build was real. Nothing the verb measured could go red on either failure, because it measured builds that returned.

## Two causes, one shape — either alone serves eager

- `Engine.compile` banks its output in torchcg's **own engine cache** (`cg-key-v1`), which it can resolve straight back out of. That is not the band adoption reads. `holes()`, `has_artifact` and every boot-time adopt address artifacts as `(cg-graph-v1, cg-env-v2)` in the `GraphStore`, and **nothing lands there unless somebody calls `publish_artifact`**. `BackgroundMint._mint_one` did. This CLI did not — it threw the child's only output away (the request has always asked for a `result` file; no caller read it).
- Nothing in the CLI ever called `put_graphs`, so adoption had no graph-set document to enumerate lanes out of and would have served eager over a correctly-filled store anyway.

## The fix: one publisher, one witness

`serving/mint.publish_compiled` is the single place a freshly-compiled artifact is put where the serving path reads it; the runtime mint and `compile` both go through it (`_manifest` came out of `BackgroundMint` with it — a compatibility set read off the artifact's own ELF was never per-mint state).

The witness is `serving_reader(cas_root)`: a `LocalGraphStore` built exactly the way `cli/daemon._adoption_source` builds it, and deliberately **not** the store this command published through. *A writer asked about its own write is not a witness.* Every build is read back through it immediately, a read-back miss is a `FAILED` specialization at the moment it happens, and the run ends with a census (`unservable`) that names the document as its own row. **The exit code is computed from that census and nothing else.**

Also: the mint child hard-coded `/usr/local/cuda`. `_ensure_cuda_home` exported the module constant even when the compose could not create it — correct only on a pod. `default_root()` now answers where THIS process can write, `Composition` carries the path it used, and `CUDA_HOME` comes from that. A benchmark harness had been carrying this as a workaround.

## Measured, on the real verb, same box and card

| | before | after |
|---|---|---|
| wall | 1561 s (26.0 min) | **140 s** |
| verdict | `built=14 (of 14)`, rc 0 | `built=14 (of 14)`, rc 0, **`SERVABLE`** |
| `torchcg/v2/artifacts` | 0 | **14** |
| `torchcg/v2/manifests` | 0 | **14** |
| serving reader sees | 0 armed / 14 holes | 14 present / 0 holes |

No rebuild was paid and **no relocation machinery was written**: `Engine.compile` resolved each existing package as a REUSE and the new publish step carried it across, so the recovery *is* the fix running.

## Red arm — 7 mutations, 7 caught

no publish · no read-back · **read-back through my own writer** · no document · census ignores the document · census ignores the artifacts · exit code computed from the outcome tally. The `PublishesNowhere` double believes its own publish and answers `has_artifact` from its own record — it is the original defect exactly, and it is what makes mutation 3 fail.

## Not fixed here

- Adoption enumerates **zero records** over the now-full store (`adopted=0, holes=0, armed=0`), so va#3's arm 2 is still blocked. Filed as **pgw#1534** — the silent branch is `AdoptSession._adopt_module`'s `if not claimed: return`, and `adopted=0/holes=0` is itself an instrument that cannot go red.
- `WorkerGraphStore.fetch_program` still raises on a miss where the `GraphStore` protocol says `Path | None` (pgw#1525's finding #1); a fix for it is UNCOMMITTED in a peer worktree and preserved as a patch — see the tracker.

Id note: filed as pgw#1532, folded into **pgw#1533** (the #1525 close-out took #1533 for the same capstone concurrently and carries the measurement body). #1532 is a burned id.

### DIFF
--- a/src/gen_worker/cli/compile.py
+++ b/src/gen_worker/cli/compile.py
@@ -18,6 +18,19 @@ than as process adoption.
 weightless trace and the mint runs against it, so ``compile`` legitimately runs
 before ``download``.
 
+**SERVABLE, OR IT FAILED (pgw#1532).** The success criterion is not "the builds
+returned". It is "the reader that arms graphs at boot can find what I wrote",
+and it is checked by asking THAT reader, through a store object this command did
+not publish through. Before this, ``compile`` reported ``built=14 (of 14)`` with
+rc=0 after 26 minutes of real inductor work and left the serving path with 14
+holes: ``Engine.compile`` banks bytes in torchcg's own engine cache, and
+NOTHING lands in the ``(cg-graph-v1, cg-env-v2)`` band adoption reads unless
+somebody calls ``publish_artifact`` — which the runtime mint did and this
+command did not. It also never called ``put_graphs``, so even a correctly placed
+artifact had no graph-set document to be found through. Both are now this
+command's job, and the read-back is what proves it: a count of builds that
+returned cannot go red on either failure, and did not.
+
 ## Address-free programs
 
 Paul ruled the exported-program blob ADDRESS-FREE: ``torch.export.save`` is
@@ -49,12 +62,18 @@ import sys
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 from . import endpoint_lock as el
 from . import work_ledger, workspace
 
 logger = logging.getLogger(__name__)
+
+#: Compile one specialization: (spec, program blob, destination) -> artifact.
+#: The production implementation is :func:`_build` (one child process per
+#: graph); an explicit one is the local seam — no GPU, no inductor — and
+#: nothing else.
+Builder = Callable[["Spec", Path, Path], Path]
 
 #: Outcome vocabulary. Closed: every specialization ends in exactly one of
 #: these and the summary counts them.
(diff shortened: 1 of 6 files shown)


PaulFidika:
## Gate state, recorded because `tests` is red

`fast gates` ✅ (the only REQUIRED context on the live ruleset — verified at enqueue time via `gh api /repos/cozy-creator/python-gen-worker/rulesets`, not from a written-down list).

`tests` ❌ with **exactly 2 failures out of 2507**, and **both are red on this PR's own base** (`2c6b1168`), reproduced locally on the canonical checkout at that exact commit:

```
FAILED tests/convert/test_flavor_token_deleted_pgw1319.py::test_the_classifier_is_deleted_and_the_vocabulary_survives
FAILED tests/convert/test_precision_class_stamp_pgw1300.py::test_the_ladder_module_no_longer_exports_a_placement
2 failed, 39 passed
```

Same two names, same cardinality, on the PR and on the base ⇒ **this branch adds zero new failures.** This diff touches **zero files under `src/gen_worker/convert/`** (`git diff origin/master --name-only | grep convert` → empty).

This is **pgw#1528** — the `convert/ladder` precision-vocabulary drift, unowned and red on master since 2026-08-19. Not absorbed here; it still needs its author.

## One thing worth reading in the log even though it merged green

This branch was cut at `3a7980a3`. pgw#1525's P0 #1 then merged onto **the same function** as #1098 (`2c6b1168`), and the stale-base diff read as *deleting* their fix. Every gate would have passed that — a revert is a valid diff. Caught by reading `git diff origin/master` before merging; rebased, and their fix is verified present afterwards. Standing rule added to `DONE-STANDARD.md`: **a green PR is not evidence you did not undo somebody.**

PR #1100, opened 2026-08-20
pgw#1534: an OOM-retry wrapper silently switched the WHOLE compiled path off

PaulFidika:
Root-caused on CPU in eight lines, fixed at the cause, and the silence that hid it is fixed too.

## The cause

`nn.Module.__call__` reads `self.forward` as an instance attribute, so installing a wrapper there is a legitimate seam this repo uses in four places. But it is not only where a compiled graph is **armed** — it is where adoption decides **whether to arm one**. torchcg claims a graph record for a `ctx.compile`-ed module when the module's forward accepts every parameter the record's ingress names, read off `inspect.signature(module.forward)`.

`oom_ladder._wrap_denoiser_forward` installed `def forward(*args, **kwargs)` onto the denoiser. That has **no named parameters at all**:

```
BEFORE install : ['encoder_hidden_states', 'return_dict', 'sample', 'timestep']
AFTER  install : []
```

So no record could ever be claimed. Not a degradation — the compiled path **off**, for that module, permanently, everywhere, and unfixable by minting because nothing was missing. pgw#1499 armed this ladder on every rung (correctly), which is what turned a per-OOM-path wrapper into a universal one.

Measured on a real boot after pgw#1533 filled the store: document with 14 graphs, **14 artifacts present** at the right `(graph, env)` positions, env identity matching — `adopted: 0, holes: 0, armed: 0`, and not one line saying why.

## The fix is the seam's rule, not this wrapper's

Anything installed onto a module's `forward` must present the signature it wrapped. `_carry_signature` does it (**not** `functools.wraps` — that copies `__dict__` and `__wrapped__` too, and this wrapper's identity should stay its own); the vae `decode` wrapper takes the same treatment so the two cannot diverge; a wrapped object with no readable signature is left alone, because an unarmed ladder never fails a load and neither may an unsigned one.

`models/lora_lifted.py` already restored `__signature__` on its own forward wrapper. The two sat four files apart and only one knew the rule.

## And the silence — the reusable half

`adopted=0, holes=0` had **three** causes and one phase:

| cause | what it means |
|---|---|
| nothing was marked | an eager endpoint, correct |
| marks matched no record | **this bug** — the compiled path is off |
| every record AMBIGUOUS | literal-twins, both disarmed; the artifacts are fine |

The last two now say so. **tcg#70** (vendored here, `d4bf7889`) records the MODULE side of an unclaimed record with the signature gap that explains it, and its message names the wrapper cause and the remedy directly. `host.py` — the path `up` and every in-process host take, where these counts were going into a span no operator reads — now WARNs with the per-module reason. `serve_adoption` gains `marks_unmatched` and `all_ambiguous` phases so the states are queryable, and `facts()` carries them.

Also fixed: `serve_adoption`'s construction-time log line reported `%d adopted, %d hole(s)` from **before** any `ctx.compile` had run, so both counts were structurally always 0 — its own `_say_outcome` docstring already explained why that had to be.

## Red arm

Removing `_carry_signature` from the denoiser wrapper turns `test_the_oom_ladder_leaves_the_denoiser_claimable` red (verified by mutation). The fence **imports** torchcg's own `_forward_parameters` rather than restating which parameter kinds count — a local restatement would pass while the real claim still failed. A second test proves the fence cannot be satisfied by deleting the wrapper.

## Ruled out cheaply, so nobody re-walks it

- **structure ambiguity** — 14 records, 14 distinct structures, 0 collisions, all concrete
- **env mismatch** — would produce 14 HOLES, not silence
- **lane-name mismatch** — contract matches exactly

Unblocks va#3 arm 2.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -696,7 +696,7 @@ torchvision = [
 # uv.lock behind, which breaks `uv sync --locked` — the first step of every CI
 # job — for every open PR; `scripts/lint_lock_pin_agreement.py` now refuses that
 # tree locally and as CI's first step.
-torchcg = { git = "https://github.com/cozy-creator/torchcg", rev = "f45fc64a9ba1323e0a44975565d822517aab8058" }
+torchcg = { git = "https://github.com/cozy-creator/torchcg", rev = "d4bf7889a72f8a7e9a47da67b0917815090cf710" }
 # Saying which pins were deleted requires naming them.
 # retired-name: historical rationale, not a live reference.
 # pgw#1310 deleted the `hashrepo` / `torch-compiled-graphs` git pins that stood
--- a/src/gen_worker/_vendor/VENDORED.toml
+++ b/src/gen_worker/_vendor/VENDORED.toml
@@ -306,7 +306,7 @@ rewrites = [
 
 [packages.torchcg]
 repo = "https://github.com/cozy-creator/torchcg"
-rev = "f45fc64a9ba1323e0a44975565d822517aab8058"
+rev = "d4bf7889a72f8a7e9a47da67b0917815090cf710"
 subdir = "src/torchcg"
 # pgw#1474 / tcg#61 RE-VENDOR (2026-08-19), from `6c91b83d`. `CompiledGraphCall`
 # stops pre-flattening the call: the loaded package is torch's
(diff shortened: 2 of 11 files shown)


PaulFidika:
Gate: `fast gates` ✅ and `drift` ✅ — `fast gates` is the only REQUIRED context on the live ruleset (read at enqueue time via `gh api .../rulesets`, not from a written-down list). No `--admin`.

**Pre-merge revert check run per DONE-STANDARD's now-required step, and it earned itself here:** this branch was cut before pgw#1537 (#1103) merged, and `git diff origin/master...HEAD` showed it deleting that refusal — which had landed five minutes earlier. Rebased onto `94331bc4`; verified present afterwards. That is the second such catch tonight and neither was visible to any gate.

Post-rebase verification, local: 47/47 across the new fence + `test_boot_end_verdict` + `test_compile_servability` + `test_oom_ladder`; `lint_lock_pin_agreement` green over the re-vendor.

PR #1103, opened 2026-08-20
pgw#1537: an unreadable endpoint module name refuses in one sentence, not a traceback

PaulFidika:
**My own regression from pgw#1533**, found by reviewing what that change made reachable rather than by anything failing.

`compile` now reads the endpoint's module name so it can publish the graph-set document under the name adoption looks it up by. That made `load_endpoint` the first thing the verb needs that can fail on the **author's** code — on a path where `run_compile` catches only `CompileError`. A broken endpoint got a bare `EndpointLoadError` traceback.

## Not fixed by swallowing it, and the distinction is the point

The neighbouring caller does swallow: `declared_floor_gb` catches everything, correctly, because an unreadable floor genuinely **is** "no floor stated" and proceeding is right.

An unreadable module **name** is not "no name". Nothing can be published under a name that could not be read, and a document published under a guess is invisible to `up` — so it refuses, naming the cause, why the name is needed, and what to do about it.

## Red arm

Removing the refusal turns `test_an_unreadable_module_name_is_a_typed_refusal_not_a_traceback` red (verified by mutation).

### DIFF
--- a/src/gen_worker/cli/compile.py
+++ b/src/gen_worker/cli/compile.py
@@ -405,12 +405,32 @@ def endpoint_module(endpoint_dir: Path) -> str:
     ``main =`` — because ``_adoption_source`` reads it the same way. Two
     spellings of "which document is this endpoint's" is exactly the drift that
     would make ``compile`` publish under a name ``up`` never asks for.
+
+    An import failure here is a TYPED refusal, not a traceback (pgw#1537).
+    pgw#1533 made this the first thing `compile` needs that can fail on the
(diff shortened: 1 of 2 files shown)


PaulFidika:
Gate: `fast gates` ✅ — the only REQUIRED context on the live ruleset (read at enqueue time via `gh api .../rulesets`). `tests` carries the same 2 pre-existing `convert/ladder` failures as pgw#1528, which this diff does not touch. Pre-merge revert check done per DONE-STANDARD's new rule: the only deletions in this diff are the two lines it replaces.

PR #1107, opened 2026-08-20
pgw#1541: the census fingerprint must OUTLIVE THE POD — one activity row for the unservable state

PaulFidika:
pgw#1541: the census fingerprint must OUTLIVE THE POD — a log line dies at teardown

Found by the H3 lane reading pgw#1536's source rather than its claim, and they
are right. That census emits logger lines ONLY: invisible to every DB-reading
harness, and DELETED AT TEARDOWN. A rental that ends in a teardown script
silently loses the exact fact the census exists to preserve, so its "free on
any run" value held only for a runner who happened to know to pull `cozy logs`
while the pod was still alive. An instrument that survives only under an
undocumented operator habit is not free; it is lucky.

WHAT LANDS

* `activity.KIND_SNAPSHOT_CENSUS` — its own kind, because it answers a
  question no other row can: whether a tree this pod holds is PINNED, and
  therefore whether the streaming engine (the only reader of pointer stubs)
  can read it at all. Verified safe on the hub side before adding it:
  `handleActivityUpdate` folds ANY non-empty kind generically into the
  activity store — no allowlist, no switch — so a new kind needs no hub
  change. Deliberately NOT reusing `snapshot_pull`, whose rows are pgw#1351's
  dedup measurement; polluting that kind would corrupt an existing query.

* ONE row, emitted ONLY when at least one tree is unservable. `step` /
  `total_steps` carry unservable/total, so the state is a NUMERIC query rather
  than a string search, and `detail` names the first few trees and the store
  root. Per-tree detail stays in the logs, where volume is free.

A FINGERPRINT, NOT A HEARTBEAT. A row per boot would be noise nobody filters,
and the second test asserts a healthy census emits NOTHING. The row exists to
mark the one combination that cannot serve — unpinned + projected — which is
the pod incident's exact state.

Red/green, $0, no GPU, no pod, on the REAL emit path (`activity.bind_sink`
bound to a recorder exactly as `arun` binds it to the worker's wire, no mock):

  the fingerprint survives as an activity row  | pass | FAIL: "exactly ONE
    (2 unservable trees -> 1 row, phase        |      | fingerprint row per
     unpinned_projected, step=2 total=2)       |      | boot … got 0"
  a HEALTHY census emits no row at all         | pass | —

The red arm restores pgw#1536's logs-only behaviour — i.e. the gap itself.

29 neighbours green; ruff clean; whole-tree mypy clean (643 files).

Follows `414f96ad` (pgw#1536). The benchmark lane's logs-while-up capture stops being the only way to read this.

### DIFF
--- a/src/gen_worker/activity.py
+++ b/src/gen_worker/activity.py
@@ -262,6 +262,22 @@ KIND_PROCESS_ROLE = "process_role"
 # ActivityUpdate field holds them and the th#1839 route serves `detail`
 # verbatim, so this needs no hub change.
 KIND_SNAPSHOT_PULL = "snapshot_pull"
+# pgw#1541: the boot SNAPSHOT CENSUS — which trees this pod holds and whether
+# each is PINNED. Its own kind because it answers a question no other row can:
+# a projected tree without its manifest pin is unreadable by the streaming
+# engine (the only reader of pointer stubs), and the pod then serves the eager
+# bridge a stubbed tree and reports `header too large` about an intact
+# checkpoint. pgw#1536 added that census as LOG LINES ONLY, which made it
+# invisible to every DB-reading harness and — worse — DELETED AT TEARDOWN: a
+# rental that ends by tearing down silently loses the exact fact the census
+# exists to preserve, and its "free on any run" value held only for a runner
+# who knew to pull `cozy logs` while the pod was alive.
+#
+# One row, emitted ONLY when at least one tree is unservable, because that is
+# the incident fingerprint and it must outlive the pod. `step`/`total_steps`
+# carry unservable/total so the state is a NUMERIC query and not a string
+# search; per-tree detail stays in the logs, where volume is free.
+KIND_SNAPSHOT_CENSUS = "snapshot_census"
 # pgw#1355: the COLD BOOT's own decomposition — per-stage spans carrying
 # start/end offsets from OS process start, plus one terminal roll-up. Its own
 # KIND because it is the only row that answers "where did this pod's cold start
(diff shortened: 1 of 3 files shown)


PaulFidika:
### Field signal: preventive-vs-post-mortem is now measured, not hypothesised

Two facts from the H3 lane, both of which I checked against this tree before repeating them:

**1. The census content DOES already reach the DB — at the price of a burned worker.**
`hub_error.py` is the post-mortem channel (`request_state.error_message_safe`), and
`serving/context.py:282` is where the stubbed-tree diagnostic gets built. But that path
only fires on a REQUEST FAILURE: the fact arrives after a worker is burned and the
placement ladder has spent its passes. sd15 has reportedly paid that price three times
tonight (6 condition-3 failures / ~6h / 3 burned pods, 4th live, all byte-identical) for
a fact this PR's boot census states ONCE, FREE, and BEFORE DISPATCH.

That is the whole argument for this PR, and it stopped being a hypothesis: the
post-mortem channel and the preventive channel report the SAME fact, and the only
difference is whether a pod dies first.

**2. Why the unconditional row (this PR's second commit) is the part that matters.**
Attribution is still pending, but the discriminating question is
predate-vs-built-this-boot, and that is read off the census at BOTH boots — absent at
boot N, present at boot N+1. A fire-on-bad-only row cannot answer it, because absence
would be ambiguous. That is exactly the bug I shipped in this PR's first commit and
fixed in its second, so the field signal arrived in time to validate the correction
rather than the original.

Not in scope here, noted for tensorhub: a blame ladder that retries a release which has
failed identically six times, on "no worker it has not already burned", is re-burning
fresh hardware against a known-poisoned release — th#2221's mark-broken use case, live.

PR #1119, opened 2026-08-20
pgw#1547: the `cell` hardcut — delete the env, move the layout, rename the wire

PaulFidika:
Paul, 2026-08-20: *"please do these renames now, fully. I want it fully renamed. You can go ahead and break everything. No legacy support, hardcut. We are pre-launch it's fine."*

Retires the 83 line-proofs the fence landed with in #1112. **Six remain, and all six are historical records** (they name the deleted env, the pre-migration path, or the exact names of tests pgw#1181 deleted) rather than deferrals.

## `GEN_WORKER_LOCAL_CELLS_DIR` is DELETED, not renamed

Paul: *"that's a configuration on where to store them, and I believe we store them in local-CAS, so you can delete this env entirely and just use whatever env we use to point to the local-CAS directory."*

`store_root()` is now `<TENSORHUB_CACHE_DIR>/compiled-graph-store` — unconditional, no branch, so the store cannot disagree with the CAS it describes. The variable leaves the runtime-env corpus (one row out, **nothing in** — a derived path is not a configuration), the config allowlist, `loader.py`, docs and fixtures.

## On disk / on the wire

`<root>/aot-cells/<ck1>/cell.tar.gz` → `<root>/graphs/<ck1>/graph.tar.gz`. No dual-read, no boot-time move.

Proto field and message **names** move with **numbers unchanged**, so the binary wire is byte-identical and the two-repo window is safe in either order (tcg#56's property). Wire **values** move with the code that emits them: `jit_cell`/`aot_cell` → `jit_graph`/`aot_graph`, boot phases `cell_*` → `graph_*`, claims `cell_read_*` → `graph_read_*`.

`reserved "requested_cell_axes"` is deliberately **not** respelled — a reserved name exists to keep that exact string un-reusable.

PROTO_DIGEST, WORKER_VALUE_CONTRACTS_DIGEST, COZY_RUNTIME_ENV_DIGEST and KEY_GRAMMAR_DIGEST all bumped; bindings regenerated with the pinned toolchain.

## Three defects this surfaced

1. `test_store_corruption_pgw1283` still reached the deleted `ENV_STORE_DIR` — caught by **mypy**, not by any test run.
2. `test_pod_isolation`'s "RELOCATED store" row asserted a scenario that no longer exists.
3. **Dropping the store's explicit privilege grant was a REAL regression.** I reasoned it was "subsumed" by the cache-root grant. That was wrong: `grant_paths` mkdirs *and* chowns each entry it is handed, and the store subdirectory does not exist on a cold pod — so only naming it gets it created owned by the compute uid. Removing it reproduced **pgw#1349 exactly** (dropped child dying on `PermissionError` at its first nested sidecar write), and only `test_pod_isolation` under a real root parent saw it. Restored, derived rather than env-read, with the parent's restated dirname now pinned to the store's own constant.

That row then failed a *second* time for a different reason — the test's fault: it computed the expected root from the **pytest** process's env while the split harness deliberately overrides `TENSORHUB_CACHE_DIR` for the child, so it probed a path nothing writes. The harness now exposes `child_env` and the row asks the child.

## Verification

mypy clean (643 files) · ruff clean · pod-isolation green under a real root parent (33 rows) · 398 passing in the affected selection · every digest/drift gate green · **a real store round-trip at the new names** — root derivation, census, verdict transitions, owed-to-sink, memo shortcut — through the module's own API, not mocks.

## Landing order

PGW LANDS FIRST per the parity registry. cozy-local and tensorhub follow and pin `peers.lock` to this merge commit.

## Recorded, not done here

The hub-side **DB schema** (`cell_store`, `cell_receipts`, `cell_revocations`, `cell_publisher_trust`, `orchestrator_cell_formulas`, `cell_key`/`cell_identity` columns, constraint names and stored enum values — 257 SQL occurrences) is a migration with a backfill and rollback story, not a rename. It stays deferred item 3 under the tracker's #1547 for Paul.


PaulFidika:
Superseded by a rebuilt branch — see the replacement PR.

**Why this one is closed rather than fixed:** its `tests` red was not the connection-refused flood in the log, and not a missed rename. Two real problems, one of which made the branch untrustworthy:

1. My torchcg **re-vendor** (for pgw#1554's corpus fix) moved the pin from `d4bf7889` to torchcg master, which crosses **tcg#71** — the precision policy gained a fourth argument. That produced `TraceLoadContext.component_dtype() takes 3 positional arguments but 4 were given` on every derive row. Adopting tcg#71 is a feature adoption, not a rename, so #1554 goes back to its own lane and the vendored rev stays where master has it.

2. Worse, and the reason for a rebuild: the branch had **silently reverted other lanes' landed master work** — `models/memory.py` (272 lines), `cli/compile.py` (143), `trace_context.py`'s tcg#71 compatibility shim, and ~17 more files across pgw#1546 / pgw#1556 / pgw#1558. A diff that quietly un-lands three other lanes cannot be patched into trustworthiness.

The replacement is rebuilt from current master with the rename re-applied from scratch, and audited so that **every deletion in the diff is vocabulary**.

PR #1152, opened 2026-08-21
pgw#1606: ctx.load_pipeline + the boot lane-selection ladder

PaulFidika:
## The audit came first, and it changed the plan

Banked in the tracker (`python-gen-worker/progress.md`, pgw#1606). Four near-misses and one active refusal.

**The headline: a multi-lane `Model` class could not boot on a pod.** `LoadedEndpoint.lane(cls, "")` raised `EndpointLoadError` for any model declaring more than one lane, and the only writers of the `contract` that would silence it are the local CLI, the daemon and `serving/__main__` — `entrypoint.py` builds `Worker(...)` with no `lane=`. `worker.py:586-588` says so in its own comment. That is why no endpoint in the fleet has ever declared two lanes, and why Paul's ratified multi-lane example had nowhere to land.

Three corrections the audit forced:

- **There is no fused triton w8a8 GEMM.** This issue's own wording said there was. The matmul is `torch._scaled_mm` only (cuBLASLt tensorwise / CUTLASS rowwise); the repo's only triton files serve nvfp4 and svdq. Built to the reality.
- **There is no fp8 artifact "gate" a worker reads.** What exists is a live micro-benchmark — a 4096³ GEMM, median of ten, requiring ≥1.10× over bf16, cached once per process. The ladder consumes that verdict and pays it once rather than growing a rival opinion.
- **The tri-state verdict is Go-only.** `tensorfs/verdict.go:177` has no pyo3 binding; its one caller is the hub's bind gate at BIND time. So the authoritative answer already exists and travels in the binding — this reads it rather than becoming a fifth copy of the pattern matcher. Ruled: tensorfs grows the binding (tensorfs#130's second item); this codes against a narrow port with the Go semantics so that lands as a one-line swap.

## What this ships

- `serving/lane_ladder.py` — ranks declared lanes using the fleet's **own** table (`models/execution_lanes`, the byte-identical twin of tensorhub's `precision/lane.go`) against card, kernel gates and staged bytes. Answers a `ResolvedLane` naming the chosen lane, the reason, and the **rejected rungs in order**, carrying pgw#1599's `DeclaredLane` verbatim so `request=`/`resident=` travel intact and `min_sm` keeps one producer.
- `ctx.load_pipeline(PipelineCls)` — the one spelling. No dtype argument, no quant if-tree, no hardware read at the author's site.
- `serving/lane_materialize.py` — materializes per the resolved contract and **proves it did**. Three refusals borrowed from torchcg's prove-it-converted discipline: converted nothing, half-converted, and a baseline lane carrying stray quantized leaves. The census reads structural markers, not dtypes — a quantized leaf reports the logical dtype it emulates, so a dtype census answers "no fp8 here" about a fully quantized model.
- The **upcast-from-fp8-download rung** becomes a rung the ladder *chooses*, with a measured saving. It already existed four times over in this tree and every one was entered because a gate failed. It declines to claim a saving it cannot measure.
- **No artifact for any lane ⇒ a priced conversion ask**, never a refusal.
- `DeployBinding` carries staged tree + bind-gate verdict + wire size **per lane contract**. A single-lane deployment carries no map and is unchanged.

## Two bugs the tests caught in the first draft

1. The baseline rung was floored out by `capability_floor_for_dtype(bf16)=80` — a statement about tensor cores, not about whether the model runs. A CUDA-less host got an empty ladder and was told to run a *conversion*: a confident wrong answer, since no conversion fixes a missing GPU.
2. "Can this card run fp8" was conflated with "are the fp8 bytes staged", which silently deleted the whole upcast saving on exactly the cards that need it.

## Proof

35 CPU tests, against fabricated card states and the **real** sdxl contract documents shipping in tensorfs today (`sdxl.diffusers-bf16@1`, `sdxl.diffusers-fp8-rowwise@1`) — so `TwoLaneModel` is a class the fleet could actually write. Both loader refusals are exercised in both directions; the confession is proven to differ between two cards, so the instrument can go red.

Adjacent existing suites green (100 passed). Two failures elsewhere (`test_quantization_lanes`, `test_serve_loop_envelope`) reproduce identically on unmodified master — they assume a CUDA-less box and this one has a 4070.

## Sequencing

Lands **after** pgw#1599 (surface + deletions + single-lane migrations); `loader._BridgedLane` is the pre-#1599 bridge and is deleted on that rebase. Multi-lane fleet declarations follow with se#816's wave. Not yet done and scoped as owed: the nvfp4 rung is proven against a fabricated document because the flat layout pgw serves has no registered contract (deliberately — it is **not** `bfl.nvfp4-preswizzled@1`, conflation measured LPIPS 1.11), and the five endpoint-side dtype/quant branches acceptance (d) names are not yet deleted.


PaulFidika:
## Update: the acceptance-(d) fence, and the GPU leg

### 🔴 Acceptance (d)'s original wording was nearly vacuous — measured, then reworded

(d) asked for *"zero dtype/quant branches in endpoint `load()` bodies"*. I built the reader, then measured the fleet before trusting it. Per endpoint, **(branches inside `load()`, branches anywhere in the package)**:

| endpoint | in `load()` | anywhere in package |
|---|---:|---:|
| `anima` | 0 | 5 |
| `joycaption` | 2 | 3 |
| `ltx-video-2.3` | 0 | 4 |
| `minimax-h3` | 0 | 8 |
| `wan-2.2` | 0 | 3 |

**21 of the 23 branches live one function outside `load()`** — minimax-h3's own `serve_recipe` module, anima's pipeline shim, ltx's and wan's helpers. **Four of those five endpoints pass a `load()`-scoped fence today with every branch intact.** A fence written to the letter of (d) would have gone green the day it landed and protected nothing: green because of its *scope*, not because of the tree, and invisible from either side once shipped.

`serving/lane_fence.py` therefore carries **both** scopes — `scan_source` (load-only, because that is still the number (d) named) and `scan_module` (every function, the real fence) — plus a fleet ledger that may only shrink and a second test asserting the module scan stays **strictly wider** than the load scan, so narrowing the fence to make it pass goes red and says why.

The reader is torch-free and AST-based, split out exactly as `load_marks_compile` was (pgw#1581/se#809): an endpoint module imports torch, so a reader reachable only through `inspect.getsource` forces every repo-side gate to write a *second* walker — and the last time that happened the two disagreed about how many endpoints compiled (2 vs 13). One reader, two entry points. Parsed, never grepped: a substring check would flag exactly the classes that documented in a comment why they do *not* branch. It also does not sweep in sdxl's FreeU branch, which is a **feature** decision the author owns — an over-broad fence gets disabled, and then it protects nothing.

Issue wording updated to "the endpoint **PACKAGE**" on the coordinator's approval.

### ✅ GPU leg — 22 seconds on an RTX 4070 Laptop

Entry and exit readings both zero compute-app rows. `scripts/pgw1606_gpu_leg.py` **refuses to run without a card** (verified card-free first) so it cannot go green proving nothing.

**The number that justifies the design:**

```
w8a8 gemm gate: mode=pertensor fp8=2.806ms bf16=4.941ms speedup=1.76x (min 1.10x)   [0.35s]
w4a4_gemm_mode() -> ''                                                              [0.00s]
```

fp8 on this card is a *benchmarked win at 1.76×*, not a hopeful default — **and 1.10× is a bar a card can miss**, which is exactly why the ladder consumes `w8a8_gemm_mode()` instead of reading an sm table. An sm table would say "sm89 ⇒ fp8 available" while knowing nothing about whether it is faster. The nvfp4 rung is vetoed in 0.00 s below sm100: cheap rejection, no benchmark paid.

**The confession, on real hardware, over the two real sdxl documents:**

```
LANE=fp8-w8a8-dynamic contract=sdxl.diffusers-fp8-rowwise@1
card=sm89/NVIDIA GeForce RTX 4070 Laptop GPU reason=gate_passed
rejected=bf16-w16a16(outranked: fp8-w8a8-dynamic ranks higher and also has bytes)
```

And a real materialization with CUDA present: built on `cuda:0`, fit rung `vae_only`, census `{'w8a8': 0, 'w4a4': 0, 'plain': 26}`, observed lane matching the resolved lane — so the assertion ran rather than being skipped.

### CI

The first run was red on both jobs and both were mine: 10 mypy-strict annotation errors in my test files, and `test_serving_adopt_first.py:197` asserting the **old** multi-lane refusal wording, which this PR legitimately moved (it now points at `resolve()` instead of being a dead end). Both fixed; `mypy src/gen_worker/serving + the new tests` is clean across 42 files, 60 targeted tests green.

PaulFidika:
## Rebased onto pgw#1599 (`d782ea5e`) — and the verdict decision, stated

### The bridge is gone

`_BridgedLane` is deleted. `resolve()` reads `model_declared_lanes(cls)` and filters to the (possibly pinned) candidate set — nothing re-parses a stamp, nothing re-derives a floor. `lane_ladder.DeclaredLane` is now an **alias** of `lane_spec.DeclaredLane` rather than the structural `Protocol` it held while this issue was built ahead of that surface: a Protocol that outlives the type it stood in for becomes a second opinion about what a lane is.

The seam matched exactly as agreed (`contract`, `contract_id`, `dtype`, `min_sm`, `spec.request`/`spec.resident`) — **nothing needed renaming.**

This lane's fixtures dropped the VRAM strings with everyone else's, and the two-lane test class now declares two **different** demand formulas (`const(GiB(1.2)) + per_mp_batch(MiB(220))` for bf16 vs `GiB(0.7)`/`MiB(140)` for fp8), which exercises the per-lane point rather than just satisfying the parser.

### 🔴 The `LaneVerdict` decision: **consume it from the binding. The hub is authoritative.** My original ask was wrong.

The audit asked tensorfs for a pyo3 `Verdict` binding so a pod could compute the tri-state locally. The #1599 lane declined to vendor it, and I want to record that the decline was correct for a stronger reason than "single implementation":

**The verdict is an ADMIT DECISION, not a lookup.** It already runs at *bind* time in the hub's bind gate against `tensorfs/verdict.go`, before the pod exists. A worker-side implementation is a **second author of admissibility**, free to admit a binding the gate would have refused — and that failure is invisible *precisely because both sides believe they agree*. Same shape as the vacuous fence in acceptance (d), and as read-as-absence.

So the authoritative answer **travels** rather than being recomputed: `lane_host.BindingVerdicts` reads the gate's own tri-state off the `DeployBinding` the hub sent. No boot round-trip, no rival matcher, one producer. Behaviour is unchanged — `BindingVerdicts` already did exactly this — but its docstring **promised a local-computation arm** once the binding landed, and that promise is now deleted and replaced with why it must not exist.

The port survives for one reason only: the ladder must be provable without a hub and without a card, and every ladder test fabricates one. It is never a place pgw decides admissibility for itself.

> A contract the binding says nothing about stays `absent`, which the ladder degrades into a priced conversion ask. That is the safe direction, and it is the whole design in one line: **pgw can ask for bytes it was not given, and cannot admit bytes it was not told about.**

The tracker's audit §6 carried a recommendation for the pyo3 binding; it is now marked **withdrawn** with this reasoning, so it does not stand as advice to a future reader. tensorfs#129's third-matcher collapse is unaffected — it was never this issue's to claim.

### State

147 tests green locally (52 in this PR, now against the real value object), mypy strict clean across 43 files.

**Still blocked on #1157** (tensorfs vendor bump, documents only): `sdxl.diffusers-nvfp4-flat@1` is not yet in the vendored `_contracts/`. On its merge I re-prove the nvfp4 arm against the real document, at which point this PR's fabricated-document tests become the red-arm pair rather than the proof — then merge on green.

PR #1160, opened 2026-08-21
pgw#1616: a serve lane declares its QUANTIZATION — vendor tensorfs 51adc50, price every ggml quant, refuse the floor-losing nvfp4 spelling

PaulFidika:
Vendors tensorfs `51adc50` (tensorfs#130 follow-up). Two documents that could not be **declared at all** now can: `sdxl.diffusers-nvfp4-flat@1` → `float4_e2m1fn`, `qwen3.6-27b-mtp.gguf-ud-q4-k-xl@1` → `q4_k`.

## The reading that kept them dtype-less was wrong, and this repo held it too

pgw#1606's fence called the nvfp4 refusal "the system working". This repo's own `test_a_dtypeless_lane_cannot_buy_a_silent_runs_anywhere_floor` said nvfp4 is block-scaled so "no single top-level dtype is true", and GGUF is a container with "no single torch dtype".

Both are true about **tensor container types** and wrong about **the field**. `dtype` names *the lane's quantization* — which is why `sdxl.diffusers-fp8-rowwise@1` declares `float8_e4m3fn` across 257 declarations of which only 36 are fp8.

**The cost was a refusal with no remedy.** Under pgw#1599's always-required `lanes=`, dtype-less means *undeclarable* — so `QwenMtpModel` could not be written and the GGUF endpoint stayed blocked, with the block attributed to a format gap (tensorfs#127) that was not in the way. That docstring now records the falsification rather than the conclusion, because the mistake is the more useful half.

## What this adds beyond the bump

**`DTYPE_MIN_SM` learns the ggml block quants at floor `0`.** Zero is correct — a block quant dequantizes on the way to the compute dtype and needs no special silicon — but it was arriving via the unknown-dtype default, and **the two zeros mean opposite things**. "Measured: no floor" and "never heard of it" are the same return value, so an absent entry prices a lane *by accident*, silently, in the permissive direction. This is the exact silent-0 sibling pgw#1606's fence was written to name — and the fence caught `q4_k` on its first real encounter, the day it landed rather than the day a GGUF lane was mis-placed.

**`FLOOR_LOSING_SPELLINGS` refuses `float4_e2m1fn_x2` by name.** torch ships `float4_e2m1fn_x2` and does *not* ship `float4_e2m1fn`, so the wrong spelling is the one that looks right, and someone will eventually "fix" the document to it. It resolves, `DTYPE_MIN_SM` does not know it, and the lane silently loses its sm100 floor and lands on Ampere. Refused rather than aliased to 100 — aliasing also works, and lets the wrong spelling spread. *A dtype that resolves through torch and drops the floor is worse than one that refuses through torch and keeps it.* The nvfp4 document asks for this in its own description; a comment cannot enforce it, this can.

**Both exemptions in the dtype fence are deleted, and neither was removed from memory.** The self-deleting waiver went red naming its own entry the moment the document grew a dtype — the entire behaviour it was built to have, observed in production rather than simulation. The GGUF exemption went because its premise was falsified upstream. No exemption remains, and a test asserts that: a new one must ship with its own expiry.

`test_the_nvfp4_lane_is_declarable_and_floors_at_blackwell` now **runs** instead of skipping, green against the real document.

## Enumerating gates

Re-pinned with the bump, each a real consequence — the contract-vectors corpus, and `rife.flownet-fp32@1` leaving the UNRATIFIED list as **the first document in the set a human ratified**. Eight remain, and the count must keep moving one at a time: a bulk drop to zero would mean the *marker* was deleted rather than the documents ratified, which is the one way that gate can be satisfied dishonestly.

## Downstream

`qwen3.6-27b-mtp-gguf` is now unblocked at the declaration surface — that endpoint's `lanes=` can name a real document. Not this PR's work, but worth knowing it moved.

188 tests green, mypy strict clean, pin-agreement gate clean.


PaulFidika:
## Correction to this PR's "Downstream" section — I checked it and it was overstated

The PR body says `qwen3.6-27b-mtp-gguf` is "now unblocked at the declaration surface". I went and read the endpoint rather than repeating the story about it, and that is **wrong in a way that matters**.

`serverless-endpoints/qwen3.6-27b-mtp-gguf/src/qwen36_27b_mtp_gguf/main.py:170-186` declares **no `lanes=` at all**. It declares `self_loading=` and `eager_only=`, and its stated reason is not the dtype:

> "llama.cpp owns the GGUF and its own kernels in a subprocess, so this process holds no torch module `ctx.compile` could mark"

So the premise that both tensorfs#130 and this repo's own test docstring carried — *"a real serve lane that an endpoint is waiting on"* — is **unverified in both repos**. No endpoint names that contract today.

What the fix actually accomplished is narrower and still worth having: **the contract is now declarable by anything that wants to name it.** Whether this particular endpoint should ever declare a lane is a separate, open question that its own header currently answers "no".

**Its real migration blocker is a different one: `eager_only=` is DELETED by pgw#1599** (`serving/model.py:730`). That header will be refused at import against current pgw regardless of any dtype. Anyone going to "unblock the GGUF endpoint" on the strength of this PR will hit that instead and lose the time — the tensorfs#127 attribution was stale, and so is the dtype attribution that replaced it.

Nothing in the diff changes; the dtype work stands on its own (the ggml quants needed explicit pricing whether or not any endpoint names them, precisely because the two zeros are indistinguishable). Only the downstream claim is corrected.

PR #1178, opened 2026-08-21
pgw#1630: the watchdog verdict is KERNEL-EVIDENCE-ONLY — the activity system demotes to labelling, and failure to measure holds

PaulFidika:
Closes pgw#1630. Follows the design review's Verdict 1 and Verdict 2 (`design-reviews/2026-08-21-watchdog.md`).

## The sharpest fact

In the pgw#1613 kills the child **passed** the kernel-evidence rung. The CAS fill kept `tree_evidence` advancing, `no_work_accrued` did not fire — and the kill came from the rung below it, `loop_wedged_no_activity`. **The parent looked at kernel-accounted proof of life and SIGKILLed anyway, because a LABEL was missing.** That contradicts `watchdog_loop`'s own docstring in the same file: *"the parent kills only what is provably NOT RUNNING; a child that runs but serves nothing is the hub's stall clock to reap."*

Two H3 pods, two pins, byte-identical death at 356 s and 687 s, exit 137, pod retired `all_declared_functions_disabled`, rental burned.

pgw#1613 opened an activity at the **one** site that had been killed. That closed a site and left the class: every future long-running path owed a ritual or died. Paul's rule 4 asks for the general solution — this is it.

## What changed

**Rung 3 is deleted.** Evidence advancing ⇒ held, unconditionally. Loop pings, declared activities and in-flight counts are reporting only; they label a stall and decide nothing. A path that forgets to declare loses a label, not a process.

**Escalation, not a cliff.** `report → diagnose → SIGTERM → SIGKILL`, one window apart, every rung requiring flatness to have *continued*. Any evidence advance resets the ladder to zero — a wedge that un-wedges mid-ladder was never a wedge.

**The diagnosis rung is new, and it is the point of having a ladder at all.** The old verdict went straight to `proc.kill()` and left nothing behind, which is why the H3 diagnosis had to be reconstructed days later from a rental invoice. `procsplit/procdiag.py` captures while the child is still ALIVE: `/proc/<pid>/status` — with **`State: D` called out by name**, since an uninterruptible wait is a stall the child could not have avoided and a SIGTERM cannot end — plus `wchan`, `syscall`, kernel stack, and a bounded `py-spy dump`. Written beside the other post-mortem artifacts so it survives a pod with no logs API. Every reader is independently guarded and the capture never raises.

**The window is derived from observation.** `W = max(floor, 3 × the largest inter-progress gap this child has actually demonstrated)`. `largest_gap_s` grows only when evidence moves *again* — a gap is only known to have been a gap once it ends — so this measures the child, not the wall clock. Both old clocks are gone:

| clock | was | now |
|---|---|---|
| `_DEFAULT_WATCHDOG_BUDGET_S = 60` | the kill budget | a `/proc` **sampling cadence** (÷4); its constant now says so |
| `_evidence_hold_window = 3 × ping cadence` | the hold window | **deleted** — a ping is a self-report, and how often a child says "still here" says nothing about how long real work may go without touching the kernel |

The floor is `Settings.watchdog_flatness_floor_s` / `GEN_WORKER_WATCHDOG_FLATNESS_FLOOR_S` — the operator lever pgw#1613 proved must exist ("NO KNOB") — and it can only ever **lengthen** a window the observed gaps have not already justified. Its default is derived and its consequence is asserted: the earliest possible kill is 4 × floor of provably-zero kernel work, and a test fails if someone shortens it under eight minutes.

**Failure to measure holds.** `no_evidence_source → KILL` inverted the burden of proof: an unreadable `/proc` became a death sentence.

> ### A defect in the first draft of this fix, recorded because it is the same shape
> Holding only when evidence had **never** been read meant that an instrument failing *after* a successful read left `flat_for(now)` growing on ELAPSED TIME while nothing was being measured — the deleted rung re-entering through the back door, and it would have SIGKILLed a healthy child on the strength of an instrument outage. The rule is now: we cannot say a child is flat when we cannot read it.
>
> Flatness accumulated *across* an outage is still real once the instrument returns, and that is sound rather than assumed: tree CPU seconds and IO bytes are monotonic, so a reading equal to the pre-outage one proves zero work happened in between.

## pgw#1613's landed scope stays

"The activity decides nothing" is not "the activity is pointless" — it is the label on `compute_child_stalled`, and the difference between a filed bug and a shrug. Its two acceptance tests are **untouched and green**. Its third asserted the deleted kill branch and is rewritten into its successor, and a new test asserts the `activity.running(KIND_BOOT_MATERIALIZE)` scope still **exists**, so a "it decides nothing, delete it" reading of this issue fails there first.

Three stale comments in `activity.py`, `boot_materialize.py` and `worker.py` described `_hang_verdict` as load-bearing for the child's LIFE; all three are corrected here. A comment describing a deleted mechanism is how the next lane resurrects it.

## Red arms — checked by breaking the fix, not by counting assertions

- restoring the cooperative rung (kill a CPU-burning child that declared nothing) turns `test_the_real_ladder_signals_NOTHING_while_evidence_advances` red — **that test is the H3 kill**;
- restoring "unreadable `/proc` is guilt" turns both unmeasurable tests red.

The structural fence reads **string literals via AST, not raw text**: a comment naming the deleted rung is documentation of why it is gone and should survive; a live string literal is the verdict being produced again.

## On the parameterized per-site test

The design asks for one per long-running site. Under Verdict 1 there is no per-site reprieve left to prove reachable, so what is proven instead is that the site's **label changes nothing**: the suite is parametrized over the four sites' real `KIND_*` plus the empty label — the site that forgot the ritual, pgw#1613's actual victim — and all five walk an identical ladder.

## Verification

- `tests/test_watchdog_evidence_pgw1630.py` — 30 passed
- `tests/test_boot_materialize.py` — green, including the rewritten pgw#1613 arm
- `mypy src/gen_worker tests` — clean, 633 files
- `ruff check src/gen_worker` — clean

No clocks were slept on anywhere in the new suite: `now` is a parameter, and every window in an assertion is computed from the track's own `window_s`.

### DIFF
--- a/changelog.d/pgw1630.md
+++ b/changelog.d/pgw1630.md
@@ -1,0 +1,66 @@
+- **pgw#1630: the compute-child watchdog kills on KERNEL EVIDENCE and on nothing else.**
+  The sharpest fact from the pgw#1613 kills is that the child *passed* the kernel-evidence
+  rung: the CAS fill kept `tree_evidence` advancing, `no_work_accrued` did not fire, and the
+  kill came from the rung below it — `loop_wedged_no_activity`. The parent looked at
+  kernel-accounted proof of life and SIGKILLed anyway, because a LABEL was missing, in direct
+  contradiction of `watchdog_loop`'s own docstring ("the parent kills only what is provably NOT
+  RUNNING"). Two H3 pods, two pins, byte-identical death at 356 s and 687 s, exit 137, pod
+  retired `all_declared_functions_disabled`, rental burned. pgw#1613 opened an activity at the
+  one site that had been killed; that closed one site and left the class, with every future
+  long-running path owing a ritual or dying. **Rung 3 is deleted.** Evidence advancing means
+  held, unconditionally. Loop pings, declared activities and in-flight counts are now REPORTING
+  ONLY — they label a stall and decide nothing, so a path that forgets to declare loses a
+  label, not a process.
+
+- **pgw#1630: escalation, not a cliff — report → diagnose → SIGTERM → SIGKILL.** The old
+  verdict went straight to `proc.kill()` and left nothing behind, which is why the H3 diagnosis
+  had to be reconstructed days later from a rental invoice. Each rung is one window past the
+  last and every rung requires flatness to have CONTINUED; any evidence advance resets the
+  ladder to zero, because a wedge that un-wedges mid-ladder was never a wedge. The new middle
+  rung captures a diagnosis while the child is still ALIVE — `/proc/<pid>/status` (with
+  `State: D` called out by name, since an uninterruptible wait is a stall the child could not
+  have avoided and a SIGTERM cannot end), `wchan`, `syscall`, the kernel stack, and a
+  bounded `py-spy dump` — written beside the other post-mortem artifacts so it survives a pod
+  with no logs API. Every reader is independently guarded and the capture never raises: a
+  diagnostic that can fail turns a stall into a crash.
+
+- **pgw#1630: the flatness window is DERIVED FROM OBSERVATION, not declared.**
+  `W = max(floor, 3 × the largest inter-progress gap this child has actually demonstrated)`.
+  A compile that legitimately ticks every 100 s earns a 300 s window; a loop that ticks every
+  200 ms is held to the floor. The largest gap only grows when evidence moves *again*, which is
+  what makes it a measurement of the child rather than of the wall clock — a gap is only known
+  to have been a gap once it ends. This satisfies the no-magic-timeouts rule rather than
+  restating it: the two clocks the old design ran on (a hardcoded 60 s budget no caller
+  overrode, and a hold window of 3× the child's self-reported PING CADENCE) are both gone, and
+  the one that remains measures absence of progress with a scale taken from the child's own
+  progress. The floor is `Settings.watchdog_flatness_floor_s`
+  (`GEN_WORKER_WATCHDOG_FLATNESS_FLOOR_S`), the operator lever pgw#1613 proved must exist, and
+  it can only ever LENGTHEN a window the observed gaps have not already justified. Its default
+  puts the earliest possible kill four windows out — eight minutes of provably-zero kernel work,
+  with a report at two and a captured diagnosis at four.
+
+- **pgw#1630: the sampling cadence and the flatness window are now SEPARATE knobs, because
+  they answer different questions.** `watchdog_budget_s` used to be both — how often the parent
+  looked, and how long silence was allowed — and collapsing the two is how the single-budget
+  cliff worked. It is now only a `/proc` sampling cadence (its constant says so), and the floor
+  under the flatness window is its own `liveness_floor_s` constructor argument on
+  `ParentControl`, taking precedence over `Settings` and then the derived default. Narrowing how
+  often the parent LOOKS must not silently narrow how long a child may legitimately be flat.
+  The integration test that reaps a SIGSTOPped child now says which of the two it means — and
+  SIGSTOP remains the honest wedge to test against, because a stopped process accrues no CPU
+  second and no byte of I/O, so it is flat by the kernel's own accounting rather than by a
+  missing label.
+
+- **pgw#1630: failure to measure HOLDS, and reports.** `no_evidence_source → KILL` inverted the
+  burden of proof: an unreadable `/proc` — a psutil import failure, a race on exit, a container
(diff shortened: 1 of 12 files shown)


PaulFidika:
**Amended `9b5e44e1` → `bf18f879`: an integration consequence the first push had wrong, and it is a real design point rather than a fixup.**

`test_wedged_child_is_killed_by_watchdog_and_pod_recovers` builds its harness with `watchdog_budget_s=3.0` and waits 30 s for the reap. Under this PR that number no longer means what the test assumed: `watchdog_budget_s` is now **only** the `/proc` sampling cadence, and the thing that decides how long flat is allowed is the floor — a derived 120 s in production, walked four times. So the child was correctly *not* killed inside 30 s, and the test failed.

The fix is not to make one number do both jobs again — collapsing them is exactly how the old single-budget cliff worked. `ParentControl` gains its own `liveness_floor_s` argument (precedence: constructor → `Settings` → derived default), the harness threads it, and the row now says which of the two it means.

SIGSTOP stays the wedge under test, and it is the honest one: a stopped process accrues no CPU second and no byte of I/O, so it is flat by the kernel's own accounting rather than by a missing label — the one case this reaper genuinely exists for.

`test_wedged_child_is_killed_by_watchdog_and_pod_recovers` — passes. `mypy` clean (634 files), `ruff` clean.

PaulFidika:
**`bf18f879` → `03aed6e0`: the second integration row that meant `watchdog_budget_s` when it now needs `liveness_floor_s`.**

`test_wedged_child_keeps_the_stream_alive_and_is_reported_not_hidden` waits for the `compute_child_stalled` dial on a SIGSTOPped child and then asserts the parent kept beating while it was frozen (`parent_beats_sent >= 4`, `beat_interval_s=1.0`).

The dial itself was already correct under the new ladder — it is now the **first rung**, flat kernel evidence past one window, three windows before anything is signalled:

```
phase=compute_child_stalled group=0 flat_s=2.1 window_s=2.0 largest_observed_gap_s=0.0
floor_s=2 slack=3.0 evidence=4.4 instrument=ok label_activity=none
label_in_flight=['r-wedge-beat#1'] label_loop_silent_s=2.4 label_ping_age_s=2.4
— measured by the parent from /proc, not self-reported by the child; the labels are
TELEMETRY and decided nothing
```

What failed was `assert 2 >= 4`: my first pass set the floor to 2 s, so the report arrived before four beats could land. The floor is exactly the quantity that row's *other* assertion is about — the interval in which the parent's beats are the only thing keeping the pod reachable — so it is now six beat-intervals, and the comment says so.

The third failure in that batch, `test_the_parent_RESPAWNS_the_census_on_a_capability_gap_pgw1436`, passes standalone and is a pure source-inspection assertion touching nothing in this PR — collateral from the two real failures in the same module run.

`-k wedged_child` → 2 passed. Full `test_procsplit.py` + `test_watchdog_evidence_pgw1630.py` + `test_boot_materialize.py` re-running locally; CI is the authority.

PaulFidika:
**Rebased onto `b2534ada` (pgw#1628's comment purge) — `03aed6e0` → `ac7f51d0`.** Seven conflicting hunks across `parent.py`, `activity.py`, `boot_materialize.py`, `config/settings.py`, `worker.py`, `test_boot_materialize.py`, `test_procsplit.py`.

Every conflict was the same shape — the purge compressed a comment this PR was rewriting — but two of them mattered beyond mechanics, because **the purge preserved statements that this PR makes false**:

- `activity.py:78` was compressed to *"The boot materialization span is load-bearing, not telemetry: the compute-child watchdog (procsplit/parent.py `_hang_verdict`) decides by what activity is OPEN…"* — a one-line summary of exactly the mechanism this PR deletes, naming a function that no longer exists.
- `boot_materialize._materialize`'s docstring carried the same claim.

Both are rewritten to the corrected constraint **in the purge's compressed style** rather than restored to my longer versions, so the purge's ruling is honoured and the statement is true. `worker.py`'s block went the other way: the purge deleted it outright, and post-#1630 there is no constraint left to state at that site, so I kept the deletion.

Resolutions favouring this PR's side were mechanical necessity, not preference: HEAD still called `_ChildSlot._hang_verdict`, which this commit deletes, so the HEAD side of those hunks could not import.

Post-rebase: `test_watchdog_evidence_pgw1630.py` + `test_boot_materialize.py` → 45 passed. `mypy` clean (635 files), `ruff` clean. The pre-rebase full local run was **107 passed, 0 failed** across `test_procsplit.py` + both suites above. CI re-running.

